PR #68, opened 2015-02-05
HPCC-JAPIs-67 Proposed solution for archived WU results

rpastrana:
Signed-off-by: rpastrana <rodrigo.pastrana@lexisnexis.com>

### DIFF
--- a/org.hpccsystems.ws.client/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
+++ b/org.hpccsystems.ws.client/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
@@ -935,9 +935,8 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
         return createAndRunWUFromECL.getWuid();
     }
 
-    /* zero based index sequence */
     /**
-     * Fetches results associated with a given WUID
+     * Fetches results associated with a given Logical name
      *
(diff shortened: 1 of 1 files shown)


drealeed:
Pretty slick. Sets off my hacky-sense, a bit...but works for me.

rpastrana:
hmmm I definitely don't want to add anything that looks like a hack...
The lookslikeaworkunit, is not great algorithm, more of a heuristic...
Perhaps a better solution would be to set both wuid and logicalname:
parameters.setWuid(wuidorlogicalname);
parameters.setLogicalName(wuidorlogicalname);
Can you verify if that will return the correct results?

drealeed:
Nope, in that case it looks for both rather than one or the other.
What about overloading the function? Keeps what we have in place, lets people figure out via the interface which one to use.
WUResultResponse result = wsWorkunits.fetchRawResults("W20140320-064321", 0, "mythor", true, 0, 50); 
WUResultResponse result = wsWorkunits.fetchRawResults( 0, "mythor", true, 0, 50,"~logicalname");

rpastrana:
@drealeed take a look at the latest commit. This way we keep the current function, but also add support for the logicalname option.
Also, the user can always create their own parameters object and call WUResultResponse fetchRawResults(WUResult parameters) directly. Let me know what you think.

drealeed:
Covers all the bases; reviewed & approved.

PR #84, opened 2015-04-24
remove overrides

### LINKED ISSUE #83: Compile errors with Java 1.8 jre
I noticed that in the DataSingleton class, there’s an @Override indicator on the run() method. This is causing a compile error with my eclipse jre (1.8). Since it’s not actually overriding anything. Can this be fixed? Same issue with Version.compareTo()—it’s implementing the compareTo() interface, not overriding a base class method.

drealeed:
Fixes #83


boneill42:
Could we get this merged?  It is currently breaking the build.

PR #86, opened 2015-04-28
Cleanup of versioning and IDE files

boneill42:
Feel free to ignore this PR, just my thoughts on versioning...

I introduced parent references so the versioning can be centrally managed.
(previously the parent used a SNAPSHOT, but the wsclient module did not, and the other submodules were on a different version)

Also, I don't believe we want IDE files checked into the code repository.  They will differ between developer's environments.


rpastrana:
@boneill42 This is a good idea, a couple of things to consider:
There are multiple projects within this repo, each can be built and used independently of the others and can therefore have their own version number.
Ultimately keeping one version for the entire project might prove to be more practical but the intent is for each of the projects to be built and released on their own.

Although this project is officially built via Maven, it is intended to be imported into Eclipse as a java project as well, there might very well be some extraneous Eclipse specific files we can trim out, but we'll keep the essential Ecipse project files in place and to be ignored by git

boneill42:
Closing -- to work on 0.6 PR.

PR #89, opened 2015-06-11
Support retrieving results, etc. from archived workunits

drealeed:
add check in getWUInfo to un-archive a workunit if it's archived before returning

add getWUResults to retrieve a workunit result

add espURL for external reference to track where a WorkunitInfo came from. (Rodrigo, there may be a better place to get this from than parsing the soap endpoint string, let me know.)


rpastrana:
A few minor comments, otherwise looks good needs some formatting.

PR #92, opened 2015-06-25
JAPI-10 ESP uses "csv" as an alias for ASCII encoding

rpastrana:
Since ESP aliases the CSV data format flag for ASCII, the client should do the same to prevent further confusion.

Also, add new method with default values for formattype=ASCII, maxsize= 8192 and rowtag = "tag"
Signed-off-by: rpastrana <rodrigo.pastrana@lexisnexis.com>


rpastrana:
@jchambers-ln this change circumvents the confusing flag issue on the ESP, please review.

jchambers-ln:
looks good -- I'll have the team test it tonight.

PR #96, opened 2015-07-22
Restructuring and removing eclipse files.

boneill42:
Eclipse files are not very portable across environments.

Instead, people wanting to use Eclipse should use:
File -> Import -> Maven -> Exising Maven Projects

That will cleanly import the project.


rpastrana:
After reviewing the changes, it appears the org.hpccsystems qualifier has been removed, unless the new nomenclature is absolutely required for Maven it should remain.

boneill42:
Squashed.

rpastrana:
Reviewed.

PR #98, opened 2015-07-24
Refactoring changes to support returning child dataset info from HPCCWsDFUClient.getDatasetFields

drealeed:
for #99, Added two new wrapper classes for the ajax-generated DFUDataColumn and DFUFileDetail classes,  DFUDataColumnInfo and DFUFileDetailInfo.

Moved the calculation of the file type & file fields into DFUFileDetailInfo.


rpastrana:
Review complete, it's hard to tell but it appears the second commit addressed all format issues.

PR #115, opened 2015-09-14
Linking modules to the parent pom.

irieksts:


Michael-Gardner:
@irieksts Morning,  Rodrigo and I were discussing this today and we decided that the way we want to handle the repository is to remove the top level JAPI pom altogether, and simply have the JAPI repository hold the subprojects.  Each with their own project pom.  Do you see any issue with this approach that we're going to take?

PR #119, opened 2015-11-17
JAPI-34 Updating generated web services from wsdl

Michael-Gardner:
@rpastrana The pom for RDF might need a wsclient version change for this to build properly with maven on your machine.  I'll rebase before a pull.  Just want to throw this up here so we can start testing it for a code review.


rdf2hpcc/src/main/java/org/hpccsystems/rdf/rdf2hpcc/RDFHPCCWsClient.java
```diff
@@ -793,48 +790,30 @@ private boolean setDropZoneInfo() throws Exception
         boolean success = false;
 
         HPCCFileSprayClient fsclient = getFileSprayClient();
-
-        DropZoneFilesRequest dropzonefilesreq = new DropZoneFilesRequest();
-
         try
         {
-            DropZoneFilesResponse dropZoneFilesResponse = fsclient.getSoapProxy().dropZoneFiles(dropzonefilesreq);
+        	targetDropzoneNetAddres = fsclient.fetchLocalDropZones()[0].getNetAddress();
```
rpastrana: are we guaranteed to have at least one netaddress?

rdf2hpcc/src/main/java/org/hpccsystems/rdf/rdf2hpcc/RDFHPCCWsClient.java
```diff
@@ -793,48 +790,30 @@ private boolean setDropZoneInfo() throws Exception
...
+            Utils.println(System.out, "Found dropzone net address: " + targetDropzoneNetAddres, false, verbosemode);
 
-            if (dropZoneFilesResponse.getExceptions() != null)
+        	targetHPCCDropzonePath = fsclient.fetchLocalDropZones()[0].getPath();
```
rpastrana: rather than making multiple calls to fsclient.fetchLocalDropZones() (line 795, 798) let's make the call once and store result in local var.

rdf2hpcc/src/main/java/org/hpccsystems/rdf/rdf2hpcc/RDFHPCCWsClient.java
```diff
@@ -793,48 +790,30 @@ private boolean setDropZoneInfo() throws Exception
...
+            Utils.println(System.out, "Found dropzone path: " + targetHPCCDropzonePath, false, verbosemode);
+            try
             {
-                Utils.println(System.out, "Failed to fetch dropzone IP address.", false, verbosemode);
+            	PhysicalFileStruct[] files = fsclient.listFiles(targetDropzoneNetAddres, targetHPCCDropzonePath, null);
+                Utils.println(System.out, "Existing Dropzone files:", true, verbosemode);
+                for (PhysicalFileStruct file : files)
```
rpastrana: Let's not go through this loop if (verbosemode == false)

rpastrana:
looks reasonable.

PR #127, opened 2016-04-05
JAPI-40 Generated and versioned

Michael-Gardner:
- Updated generated wsdl stub code
- Added wsesdlconfig
- Versioned to 5.6.0
- Latest = 6.0.0

Signed-off-by: Michael Gardner <michael.gardner@lexisnexis.com>


wsclient/.settings/org.eclipse.core.resources.prefs
```diff
@@ -1,4 +1,7 @@
 eclipse.preferences.version=1
```
rpastrana: add this file to the git ignore list

wsclient/.settings/org.eclipse.jdt.core.prefs
```diff
@@ -1,11 +1,12 @@
 eclipse.preferences.version=1
```
rpastrana: git ignore

Michael-Gardner:
@rpastrana 

- Cleaned up the .gitignore files.  Added in .settings/ directory and removed the wsclient/.settings/ files from the git cache so they should no longer show up when you do a commit or git status.
- Fixed test files for new versions of wsclient
- Changes to client files to fix WuActions -> ECLWUActions change
- Changes to client files for ApplicationValues instead of each individual name.  I kept the function signature the same though (it accepts appKey, appName, appData which get translated to application, name, value respectively.)  Thinking about adding a second signature that takes an ApplicationValues[] array as an argument.  Ideas?

.gitignore
```diff
@@ -1,2 +1,10 @@
 */target
 bin/
+.project
+.settings/
+EclRecord.tokens
```
rpastrana: good idea

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -700,9 +673,8 @@ public WUQueryResponse workUnitUQuery(String wuid, String jobname, String cluste
             if (jobname != null) request.setJobname(jobname);
             if (count != null) request.setCount(count);
             if (owner != null) request.setOwner(owner);
-            if (appKey != null) request.setApplicationKey(appKey);
-            if (appData != null) request.setApplicationData(appData);
-            if (appName != null) request.setApplicationName(appName);
+            ApplicationValue applicationValue[] = {new ApplicationValue(appKey,appName,appData)};
```
rpastrana: do we have a version that accepts appvalues ?

Michael-Gardner: Not at the moment.  I was thinking of adding a signature in that would replace the appName, appKey, appData parameters with an ApplicationValue[] parameter.  I'd use the code from the current implementation and then just make the signature we have currently (for backwards compatibility) be a wrapper to the new sig.

rpastrana: Let's add that new wrapper now.

rpastrana:
looks reasonable.

PR #128, opened 2016-04-08
Check for errors in submitted workunit before running

drealeed:
Fixes JAPI-42. Take a look and let me know if this is the right way to fix it.

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
@@ -888,6 +888,17 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
             createdWU.setCluster(wu.getCluster());
             submitWU(createdWU); // if no exception proceed
             this.monitorWUToCompletion(createdWU);
+
+            //exceptions, etc. aren't always included in the submit response; do another request to get all workunit info
+            WUInfo parameters=new WUInfo();
+            parameters.setWuid(createdWU.getWuid());
+            parameters.setIncludeExceptions(true);            
+            WUInfoResponse res=wsWorkunitsServiceSoapProxy.WUInfo(parameters);            
+            if (createdWU.getExceptions() == null 
+            		&& res.getWorkunit() != null 
+            		&& res.getWorkunit().getExceptions() != null) {
+            	this.throwWsWUExceptions(res.getWorkunit().getExceptions(),"Workunit Compile Failed");
+            } 
         }
         return createdWU;
     }
@@ -954,7 +965,7 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
 
             if (compiledWU != null)
             {
-                WURun runparameters = new WURun();
+            	WURun runparameters = new WURun();
                 runparameters.setWuid(compiledWU.getWuid());
                 runparameters.setVariables(wu.getNamedValues());
                 runparameters.setCluster(wu.getCluster());
@@ -1141,6 +1152,7 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
             return wsWorkunitsServiceSoapProxy.WUResult(parameters);
     }
 
+   
     /**
      * Creates and throws exception with exception message response from WS
      *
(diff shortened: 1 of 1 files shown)


wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -888,6 +888,17 @@ public WorkunitInfo compileWUFromECL(WorkunitInfo wu ) throws Exception
             createdWU.setCluster(wu.getCluster());
             submitWU(createdWU); // if no exception proceed
             this.monitorWUToCompletion(createdWU);
+
+            //exceptions, etc. aren't always included in the submit response; do another request to get all workunit info
+            WUInfo parameters=new WUInfo();
```
rpastrana: I wish we could build this into monitorWUToCompletion since that method is already calling WUInfo

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -888,6 +888,17 @@ public WorkunitInfo compileWUFromECL(WorkunitInfo wu ) throws Exception
             createdWU.setCluster(wu.getCluster());
             submitWU(createdWU); // if no exception proceed
             this.monitorWUToCompletion(createdWU);
+
+            //exceptions, etc. aren't always included in the submit response; do another request to get all workunit info
```
rpastrana: We should prob. make use of WorkunitInfo getWUInfo(String wuid,boolean unarchive) here

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -1141,6 +1152,7 @@ public WUResultResponse fetchRawResults(WUResult parameters) throws Exception
             return wsWorkunitsServiceSoapProxy.WUResult(parameters);
     }
 
+
```
rpastrana: let's remove this line

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -954,7 +965,7 @@ private WURunResponse createAndRunWUFromECL(WorkunitInfo wu) throws Exception
 
             if (compiledWU != null)
             {
-                WURun runparameters = new WURun();
+            	WURun runparameters = new WURun();
```
rpastrana: indenting might be a little off

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -1162,6 +1174,30 @@ private void throwWsWUExceptions(ArrayOfEspException wsWUResponseExceptions, Str
...
     }
 
+    /**
+     * Creates and throws exception with exception message response from WS
+     *
+     * @param eclExceptions - the array of ECLException objects to throw
+     * @param message - the prefix message
+     * @throws Exception
+     */
+    private void throwWsWUExceptions(ECLException[] eclExceptions, String message) throws Exception
```
rpastrana: Let's distinguish this method from the existing throwWsWUExceptions by labeling it throwWUECLExceptions

rpastrana:
@drealeed reviewed, left a few minor comments.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
drealeed: That would be fantastic. Currently hipie has to catch exceptions returning from createAndRunWuFromEcl, check the message to see if it's a timeout, and then make a call to WUInfo, passing in the jobname (since no workunit ID is returned) to try and get the actual status of the workunit. Perhaps instead of throwing an error, on timeout, a WorkunitInfo object could be returned containing the wuid + the timeout error?

drealeed:
Made all recommended changes

richardkchapman:
@rpastrana Back to you

rpastrana:
looks good.

rpastrana:
@Michael-Gardner

PR #130, opened 2016-04-19
Filter to actual errors

drealeed:
Don't fail if only warnings exist in the Workunit eclexceptions collection.

For https://track.hpccsystems.com/browse/JAPI-44


Michael-Gardner:
Looks good

PR #135, opened 2016-06-01
Flesh out workunit info cloning--there were some methods in .58 not being cloned

drealeed:
Also put them in alphabetical order to make it easier to track missing methods going forwards.

Jira issue https://track.hpccsystems.com/browse/JAPI-48


Michael-Gardner:
Looks good

PR #137, opened 2016-06-15
Add annotations to fields, not records.

lrenn:
Let me know if you want this PR to a different branch.

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecordReader.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecordReader.java
@@ -442,7 +442,10 @@ public class EclRecordReader extends EclRecordBaseListener
     @Override
     public void exitAnnotation(final EclRecordParser.AnnotationContext ctx) {
         super.exitAnnotation(ctx);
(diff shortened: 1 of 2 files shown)


wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecordReader.java
```diff
@@ -442,7 +442,10 @@ public void enterComment(final EclRecordParser.CommentContext ctx) {
     @Override
     public void exitAnnotation(final EclRecordParser.AnnotationContext ctx) {
         super.exitAnnotation(ctx);
-        currentrec.getAnnotations().add(new DFUDataColumnAnnotation(annotationName, annotationParams));
+        if (currentrec.getChildColumns().size() > 0) {
+            final DFUDataColumnInfo info = currentrec.getChildColumns().get(currentrec.getChildColumns().size() - 1);
```
Michael-Gardner: Should this really be 'final' ?

lrenn: I don't change the reference, so IMHO, yes.  If you look at some of my code, I always use final unless I'm going to change a reference (which I try to avoid doing).

lrenn:
Just wondering around when I a build with this might be available.

PR #147, opened 2017-01-03
JAPI-71 Readme change and update of source for wsclient-1.0.0 dependency

Michael-Gardner:
@rpastrana I'd rather not copy over the rdf file due to copy write concerns.  I'm searched around but I can't find a license file to go with that document, and if there is no document then generally the publisher retains all copy write protections.  So linking to it is safer.

I had to update the imports because of refactoring of wsclient back in 0.6.3.  I also had to update the filespray imports, although in the future it would probably be best if we got rid of these imports and handled everything directly through the API we have available in wsclient and not through the wsdl generated code.

Signed-off-by: Michael Gardner <michael.gardner@lexisnexis.com>


rpastrana:
@Michael-Gardner looks good, let's open an issue for removing use of generated stub code from the rdf tool.

PR #161, opened 2017-07-31
JAPI-00083: StateID from State only.

lrenn:
Perhaps not the best fix, but if for some reason the WS doesn’t return the state ID, but does return the state name, allow these methods to work and not throw an NPE.  This previously happened when checking an archived work unit.

I also considered setting the ID during the copy() method of WorkUnitInfo but then the state of the object wouldn't represent the true response.

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Via the attached JUnit tests.

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
@@ -382,7 +382,7 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
         if (wu != null) {
             if (wu.getStateID() != null) {
                 return getStateID(wu.getStateID());
-            } else if (wu.getState() != null && wu.getState().trim().length() != 0) {
+            } else if (wu.getState() != null) {
                 return Workunit.translateWUState(wu.getState());
             }
         }
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
@@ -56,6 +56,7 @@ import org.hpccsystems.ws.client.gen.wsworkunits.v1_62.WsWorkunitsServiceSoap;
 import org.hpccsystems.ws.client.gen.wsworkunits.v1_62.WsWorkunitsServiceSoapProxy;
 import org.hpccsystems.ws.client.gen.wsworkunits.v1_62.ThorLogInfo;
 import org.hpccsystems.ws.client.platform.WUState;
+import org.hpccsystems.ws.client.platform.Workunit;
 import org.hpccsystems.ws.client.platform.WorkunitInfo;
 import org.hpccsystems.ws.client.platform.Version;
 import org.hpccsystems.ws.client.utils.Connection;
@@ -378,10 +379,15 @@ public class HPCCWsWorkUnitsClient extends DataSingleton
      */
     public static WUState getStateID(WorkunitInfo wu)
     {
-        if (wu != null)
-            return getStateID(wu.getStateID());
-        else
-            return WUState.UNKNOWN;
+        if (wu != null) {
+            if (wu.getStateID() != null) {
+                return getStateID(wu.getStateID());
+            } else if (wu.getState() != null && wu.getState().trim().length() != 0) {
+                return Workunit.translateWUState(wu.getState());
+            }
+        }
+
+        return WUState.UNKNOWN;
     }
 
     /**
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/platform/Workunit.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/platform/Workunit.java
@@ -156,7 +156,7 @@ public class Workunit extends DataSingleton
         return WUState.UNKNOWN;
     }
 
-    public static boolean isFailedState(String state) throws Exception
+    public static boolean isFailedState(String state)
     {
         WUState statecode = translateWUState(state);
         switch (statecode)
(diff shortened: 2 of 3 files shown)


rpastrana:
@lrenn can you open an issue against HPCC for this issue: "if for some reason the WS doesn’t return the state ID, but does return the state name..."

wsclient/src/main/java/org/hpccsystems/ws/client/platform/Workunit.java
```diff
@@ -192,11 +192,8 @@ public static boolean isFailedState(String state) throws Exception
             WuStateNameMap.put("RUNNING", WUState.RUNNING);
     }
 
-    public static WUState translateWUState(String state) throws Exception
+    public static WUState translateWUState(String state)
     {
-        if (WuStateNameMap.size() <= 0)
```
rpastrana: why was this block removed? If the static block above executes, we're never going to throw this exception. But if the code changes, this exception will let us quickly identify the issue.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -378,10 +379,15 @@ static public boolean isWorkunitComplete(WUState state)
...
     public static WUState getStateID(WorkunitInfo wu)
     {
-        if (wu != null)
-            return getStateID(wu.getStateID());
-        else
-            return WUState.UNKNOWN;
+        if (wu != null) {
+            if (wu.getStateID() != null) {
+                return getStateID(wu.getStateID());
+            } else if (wu.getState() != null && wu.getState().trim().length() != 0) {
```
rpastrana: we should either trim in translateWUState() or make sure the argument in the translateWUstate call below is trimmed as well. As is, we might pass this condition (wu.getState().trim().length() != 0), but fail to translate due to leading or trailing spaces on an otherwise valid state string.

wsclient/src/test/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClientTest.java
```diff
@@ -0,0 +1,39 @@
...
+import org.hpccsystems.ws.client.platform.WUState;
+import org.hpccsystems.ws.client.platform.Workunit;
+import org.hpccsystems.ws.client.platform.WorkunitInfo;
+import org.junit.Test;
+
+import static org.junit.Assert.*;
+
+
+public class HPCCWsWorkUnitsClientTest {
+    @Test
```
rpastrana: these are good, can we add comments so we know what each test is doing?

wsclient/src/main/java/org/hpccsystems/ws/client/platform/Workunit.java
lrenn: I didn't think it was worth throwing an exception to catch a static block not executing, that's defined right above.  I would have then had to handle the exception by swallowing, returning UNKNOWN, or changing the sig to also throw.

lrenn: I'll add to my test case to make sure the map isn't empty, but throwing doesn't make sense.

lrenn: Actually, the tests I added already ensure the map isn't empty.

rpastrana: ok, but the throws was removed, which could affect code that called this function, please make sure that callers of this function (and callers of the caller) are not throwing this exception anymore.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -382,7 +382,7 @@ public static WUState getStateID(WorkunitInfo wu)
         if (wu != null) {
             if (wu.getStateID() != null) {
                 return getStateID(wu.getStateID());
-            } else if (wu.getState() != null && wu.getState().trim().length() != 0) {
+            } else if (wu.getState() != null) {
```
rpastrana: ok, so now it looks like you were only concerned about attempting to translate a string made up of white space, is that no longer a concern? Also, are valid states with leading, trailing white space not a concern? If they are, let's trim in the translate function.

lrenn: I'm not concerned about a string made up of only whitespace.  It will return UNKNOWN just as if it was empty.  Adding the trim().length() check was my original mistake.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
```diff
@@ -378,10 +379,15 @@ static public boolean isWorkunitComplete(WUState state)
      */
     public static WUState getStateID(WorkunitInfo wu)
     {
-        if (wu != null)
-            return getStateID(wu.getStateID());
-        else
-            return WUState.UNKNOWN;
+        if (wu != null) {
+            if (wu.getStateID() != null) {
```
rpastrana: @lrenn please either remove the curly brackets around the if and else if blocks, or move the open curly to a new line as well as the else if, and squash your commits into 1 so we can finalize this pull request. Thanks.

lrenn: http://www.oracle.com/technetwork/java/javase/documentation/codeconventions-142311.html#449

https://google.github.io/styleguide/javaguide.html#s4.1-braces

I'll change the brace format, but most Java codebases use those standards.  As for squashing, that can be done from github when you merge, but I'll close this PR and submit a new one.

lrenn:
Closing to squash the commits.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsWorkUnitsClient.java
rpastrana: @lrenn as part of the Pull Request process you checked off the "My code follows the code style of this project. " but I noticed " I have applied the Eclipse code-format template provided." was not checked off. 
Please take a look at this link: https://github.com/hpcc-systems/HPCC-JAPIs#contributions

PR #195, opened 2018-07-10
Japi-120: fixes to get file metadata

drealeed:
## Type of change:

## Checklist:
  - [X] I have performed system test and covered possible regressions and side effects.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization
  - [X] This change fixes the problem, not just the symptom
  - [X] The target branch of this pull request is appropriate for such a change.

This updates the getFileMetadata method:
* when requesting metadata for a superfile, the metadata for the first subfile is returned instead. If there are no subfiles in the superfile an error is thrown describing this.
* if record ecl exists it is parsed to extract additional metadata info (xpath, maxlength, etc) which are appended to the metadata columns.
* the method now returns List<DFUDataColumnInfo> rather than DFUDataColumn[] in order to return the extra metadata.

I discovered during testing that " new DFUDataColumnInfo(DFUDataColumn base) " wasn't copying the child columns of the DFUDataColumn to DFUDataColumnInfo, and fixed this.

I also removed DFUDataColumnInfo's dependency on a soap-generated specific DFUDataColumn version.
 
## Testing:
* added a test to RampsDevRegressionTest that exercises the new logic in getFileMetaData. I created specific files for the test on the cluster so (knock on wood) they won't be deleted by anyone else.
* ran mvn test from the command line.
* reran the ecl parsing regression test, which calls getFileMetaData, against ramps_dev and confirmed that no unexpected or new errors occurred.


rpastrana:
@drealeed it looks like your branch might be based off of master, we might need to rebase on wsclient-1.4.0

drealeed:
It's definitely based off of 1.4.0--I actually branched it off of the JAPI 114 branch. What about the changes makes you think it's based off of master?

rpastrana:
Are all 5 of the commits on this branch intended to be in this PR?

rpastrana:
@drealeed  I think this just needs a rebase of wsclient-1.4.0 onto your  japi-120-getFileMetadata branch.

drealeed:
@rpastrana , what happened is the following:
1. create japi-114 branch, make changes, create pull request
2. branch japi-120 off of japi-114, make metadata changes
3. squash the two commits on the japi-114 branch for the pull request, and merge it to wsclient-1.4.0
4. rebase japi-120 branch off of wsclient-1.4.0, and put in a pr for it

Because I squashed the two japi-114 commits after branching japi-120 off of the japi-114 branch, even when rebasing japi-120 off of wsclient-140, github doesn't recognize the squashed commit in japi-114 as the same as the two original commits in japi-120 that the squash was based off of. Therefore it lists them in the pr. However, the changes in those commits _have_ been merged to wsclient-1.4.0, so when I look at the file changes, the 114 changes aren't there.

 My bad for not branching japi-120 off of wsclient-1.4.0, or waiting until japi-114 was pulled before creating the japi-120 branch. If you squash the commits for this PR when merging, the problem will resolve itself.

I did check and the 114 changes aren't in this pr.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -14,6 +14,8 @@
 import javax.xml.parsers.DocumentBuilderFactory;
 
 import org.apache.axis.client.Stub;
+import org.apache.commons.lang3.StringUtils;
```
rpastrana: is this found in "stock" SDK, or is it a new dependency?

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+     * Use this function to retrieve file metadata such as column information,
+     * for superfiles the metadata from the first subfile will be returned.
      * @param logicalname    - Logical filename.
      * @param clustername    - Optional - The cluster the logical filename is associated with.
-     * @return Array of DFUDataColumns
+     * @return ArrayList of DFUDataColumnInfo
      * @throws Exception
      */
-    public DFUDataColumn[] getFileMetaData(String logicalname, String clustername) throws Exception
+    public List<DFUDataColumnInfo> getFileMetaData(String logicalname, String clustername) throws Exception
```
rpastrana: the new signature makes sense, but I'm concerned about changing an interface like this, specially in a minor release. Does it make sense to overload the method w/ both signatures 
public List<DFUDataColumnInfo> getFileMetaData(String logicalname, String clustername) throws Exception
and
public DFUDataColumn[] getFileMetaData(String logicalname, String clustername) throws Exception

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
-        DFUDataColumn[] cols = null;
-
-        req.setLogicalFileName(logicalname);
-
-        if (clustername != null)
-            req.setClusterName(clustername);
 
+        List<DFUDataColumnInfo> cols=new ArrayList<DFUDataColumnInfo>();
+        String eclrecord=null;
+        //getFileMetadata fails for superfiles; use first subfile to retrieve record structure if this is the case
```
rpastrana: There might be legacy superfiles w/ multiple subfiles with different rec defs, we might want to interrogate all subfiles and confirm they are homogeneous. If not, throw exception?

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+        //also retrieve ecl to extract extra information (xpath, maxlength) not in getFileMetadata
         try
         {
-            DFUGetFileMetaDataResponse resp = proxy.DFUGetFileMetaData(req);
-            if (resp == null)
-                return cols;
+            DFUInfoResponse details=this.getFileInfo(logicalname, clustername);
+            if (details != null && details.getFileDetail() != null) {
```
rpastrana: let's avoid the K&R format (curly braces should be in new line)

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+                eclrecord=details.getFileDetail().getEcl();
+                    if (details.getFileDetail().getIsSuperfile()) {
```
rpastrana: indenting here seems off

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+                        } else {
+                            throw new Exception(logicalname + " is a superfile with no subfiles, cannot determine file structure");
+                        }
+                }
+            }
+        }
+        catch (Exception e)
+        {
+            String msg="Error calling DFUInfo for " + logicalname + ":" + e.getMessage();
+            if (verbose)
```
rpastrana: I don't think we need this check, 
 Utils.println(System.out, msg, true, verbose); <-- this will handle the if (Verbose)

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -260,34 +263,67 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+            {
+                    Utils.println(System.out, msg, false, true);
+            }
+            throw new Exception (msg,e);
+        }
 
-            this.handleException(resp.getExceptions());
+        try {
```
rpastrana: { in new line here and a few other places

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -342,7 +408,9 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
                     DFUDataColumn[] thesecols = (DFUDataColumn[]) r;
                     if (r != null)
                     {
-                        cols.addAll(Arrays.asList(thesecols));
+                        for (DFUDataColumn col:Arrays.asList(thesecols)) {
```
rpastrana: line 409 should really be if (thesecols != null)

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -351,7 +419,9 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
```
rpastrana: same thing in line 420, if (thesecols != null), it's equivalent, but we're actually referencing thesecols in the loop

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
```diff
@@ -3,15 +3,19 @@
 import java.util.ArrayList;
 import java.util.List;
 
-import org.hpccsystems.ws.client.gen.wsdfu.v1_36.DFUDataColumn;
-
 // This class wraps the generated soap ECL Workunit, providing comparable and to-string methods for end-users.
-public class DFUDataColumnInfo extends DFUDataColumn
+public class DFUDataColumnInfo
```
rpastrana: why do we no longer extend DFUDataColumn?

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
```diff
@@ -3,15 +3,19 @@
...
 {
-    /**
-     * 
-     */
-    private static final long       serialVersionUID = 1L;
```
rpastrana: I can't tell, but if we serialize this class, we might want to keep serialVersionUID, and update it

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
```diff
@@ -26,11 +30,22 @@
      * 
      * @param base
      */
-    public DFUDataColumnInfo(DFUDataColumn base)
+    public DFUDataColumnInfo(org.hpccsystems.ws.client.gen.wsdfu.v1_36.DFUDataColumn base)
```
rpastrana: I'm a little concerned about this... we can't create a DFUDataColumnInfo based on post wsdfu1.36 DFUDataColumns?

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
```diff
@@ -76,7 +91,7 @@ public String toString()
      * 
      * @param base
      */
-    protected void copy(DFUDataColumn base)
+    protected void copy(org.hpccsystems.ws.client.gen.wsdfu.v1_36.DFUDataColumn base)
```
rpastrana: are we losing the ability to copy pre/post 1.36 DFUDatacolumns ? do we need other copy methods which accept other versions of DFUDataColumn ?

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
```diff
@@ -92,8 +107,46 @@ protected void copy(DFUDataColumn base)
         this.setIsKeyedColumn(base.getIsKeyedColumn());
         this.setIsNaturalColumn(base.getIsNaturalColumn());
         this.setMaxSize(base.getMaxSize());
+        if (base.getDataColumns() != null) {
```
rpastrana: is this only available in 1.36?

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
drealeed: It's already in the pom file for wsclient.

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
drealeed: It's not needed as a top level class now that it's no longer extending DFUDataColumn, it has its own serial uid generated

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
drealeed: I checked the call heirarchy for this method and currently it's only called by wsclient getFileDetails() and two unit tests in wsclient. It's not used in HIPIE, DSP or any other ln products I'm aware of.

It is a bit draconian to change the interface in a minor release. We can't have the same method signature returning two different objects though, so a safer bet would be to add a  getFileMetadataInfo method returning the new list.  I'll add a deprecated flag to the old method.

wsclient/src/main/java/org/hpccsystems/ws/client/platform/DFUDataColumnInfo.java
drealeed: nope, it's available in all

drealeed:
Ran HPCCWsDFUClient through the formatter.
Created a datacolumns var to reference.
Not sure what we can do for heterogynous superfiles. Is it worth cycling through all of them, and throwing an error if any ecl mismatch is found?

rpastrana:
@drealeed I think that's exactly what we should do. 99% of the time, they'll be equivalent, but we'll have our bases covered for that one-off superfile out there

drealeed:
@rpastrana I'll add this in and test it.

Keep in mind that the getfiledata soap call needed to retrieve the ecl takes .5 to 1 second per call currently, so a superfile with > 20 files will take a long time to return metadata.

rpastrana:
@drealeed based on the overhead you mention, it doesn't seem worthwhile. This feature will be handled by the ESP in 7.0.0 anyway.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -263,6 +263,35 @@ public NodeList getFileData(String logicalname, Long beginrow, Integer numrows,
...
+        }
+        for (int i = 0; i < subfiles.length; i++)
+        {
+            DFUInfoResponse details = this.getFileInfo(subfiles[i], cluster);
+            if (details == null || details.getFileDetail() == null)
+            {
+                continue;
+            }
+            String thisecl = details.getFileDetail().getEcl() == null ? "" : details.getFileDetail().getEcl();
+            if (eclrecord == null)
```
rpastrana: if (eclrecord == null && thisecl.len != 0 )  ?

drealeed: no, this should only be populated once by the first record in the list. If some file's added to a superfile with an empty record structure, we still want to check that against other records.

PR #277, opened 2019-04-23
JAPI-131 standard error classes

drealeed:
This is the initial check in for standard error classes for use by the hpcc-japi suite of projects.

I based it on the common error class used by salt, hipie, kel, etc.

These basic implementations of the error classes, error codes, etc can also be extended/implemented by dfu and wsclient specific errors with dfu/wsclient specific error types and error codes, if so desired.

Once these classes exist, an errorblock of errors/warnings/etc. can be returned from a method rather than having to throw an error. This is very useful in cases where multiple errors can be created by a method (when compiling ecl, for example).

The questions and issues raised by Rodrigo in the earlier PR have been addressed.

I also updated the java compilation version to 1.8 in the pom, and added in configuration to allow testing of integration junit tests when desired.

## Type of change:

## Checklist:
  - [X] I have applied the Eclipse code-format template provided.
  - [X] Any new interfaces or exported functions are appropriately commented.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorBlock.java
```diff
@@ -0,0 +1,154 @@
...
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see java.lang.Object#toString()
+     */
+    @Override
+    public String toString()
+    {
+        StringBuilder s = new StringBuilder();
```
rpastrana: consider synchronizing

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorBlock.java
```diff
@@ -0,0 +1,154 @@
...
+        for (HpccError e : getErrors())
+        {
+            sb.append(e.toEclIdeString());
+        }
+        return sb.toString();
+    }
+
+    public void clear()
+    {
+        he.clear();
```
rpastrana: should this be sychronized?

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorCode.java
```diff
@@ -0,0 +1,20 @@
+package org.hpccsystems.commons.errors;
+
+public enum HpccErrorCode implements IErrorCode
+{
+
+    UNCAUGHT_EXCEPTION(200), INFO(300);
```
rpastrana: this list of codes feels incomplete. Do we know which hpcc errors we're representing here so we can get the full list from the hpcc code?
for example dfu error codes: https://github.com/hpcc-systems/HPCC-Platform/blob/ff2b7df781c6205b12795e131751b1eecf14fbc4/dali/dfu/dfuerror.hpp

rpastrana: Also, info sounds more like a err level

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorLevel.java
```diff
@@ -0,0 +1,32 @@
...
+ */
+public enum HpccErrorLevel
+{
+    FATAL, ERROR, WARNING, INFO;
+
+    public static HpccErrorLevel getErrorLevel(int eclenum)
+    {
+        if (eclenum == 0)
+        {
+            return null;
```
rpastrana: what does null represent in this case? no error? unknown error level?
also wha about eclenum < 0 ?

commons-hpcc/pom.xml
```diff
@@ -64,8 +64,8 @@
                 <artifactId>maven-compiler-plugin</artifactId>
                 <version>3.1</version>
                 <configuration>
-                    <source>1.6</source>
-                    <target>1.6</target>
+                    <source>1.8</source>
```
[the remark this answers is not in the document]
drealeed: all java.nio class usage, and lambda expressions

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccError.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    /*
+     * (non-Javadoc)
+     * 
+     * @see java.lang.Object#hashCode()
+     */
+    public int hashCode()
+    {
+        return this.toEclIdeString().hashCode();
```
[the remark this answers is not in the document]
drealeed: This was copied over from the tardis ErrorBlock code--in that codebase, we overrode hashcode to prevent completely duplicate errors from existing multiple times in an errorblock. (If two errors had the same tostring, the are considered "identical". ) In this case, if the tostring for two errors both have "unknown file"  + the same error message they'd be the same hashcode; but if their error message was different it'd be different.

The hashCode override to ensure error uniqueness in an error block is tbh a bit of a hack and I understand if you'd rather have it removed from this codebase.

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorBlock.java
drealeed: Not necessary as he is a copyOnWriteArraySet (see https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CopyOnWriteArraySet.html for details)

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorCode.java
drealeed: This is a very small set. The intention was that wsclient and dfsclient would implement IErrorCode with their own wsclient/dfsclient-specific error codes; this class is here as an example/base class to extend.

drealeed: An HpccError object takes an error level, error type and error code. In the case of info error level messages I have a basic "info" error code to use.

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorBlock.java
drealeed: Not necessary as the internal array  is a copyOnWriteArraySet (see https://docs.oracle.com/javase/8/docs/api/java/util/concurrent/CopyOnWriteArraySet.html for details)

commons-hpcc/src/main/java/org/hpccsystems/commons/errors/HpccErrorLevel.java
drealeed: The error code in the soap-generated ecl error objects is an int, so 0=null.

PR #323, opened 2019-10-22
JAPI-245 Top Level Pom Deployment Fix

Michael-Gardner:
Signed-off-by: Michael Gardner <michael.gardner@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/commons-hpcc/pom.xml
+++ b/commons-hpcc/pom.xml
@@ -8,7 +8,7 @@
 
   <parent>
     <groupId>org.hpccsystems</groupId>
-    <artifactId>HPCC-JAPIs</artifactId>
+    <artifactId>japi</artifactId>
     <version>7.6.4-SNAPSHOT</version>
   </parent>
 
--- a/dfsclient/pom.xml
+++ b/dfsclient/pom.xml
@@ -8,7 +8,7 @@
 
   <parent>
     <groupId>org.hpccsystems</groupId>
-    <artifactId>HPCC-JAPIs</artifactId>
+    <artifactId>japi</artifactId>
     <version>7.6.4-SNAPSHOT</version>
   </parent>
 
--- a/pom.xml
+++ b/pom.xml
@@ -3,11 +3,12 @@
 <project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.hpccsystems</groupId>
-  <artifactId>HPCC-JAPIs</artifactId>
+  <artifactId>japi</artifactId>
   <version>7.6.4-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>HPCC Systems Java Projects</name>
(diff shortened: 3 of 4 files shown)


Michael-Gardner:
@rpastrana @drealeed

So the top level pom isn't getting published correctly with 7.6.0 and beyond.  For the 7.6.4 release I'd like to get this fixed.  The changes I've made will allow us a directory structure like this;

org
- hpccsystems
-- <version dir>
--- pom.xml
-- wsclient
--- <version dir>
---- jars and poms for wsclient
-- etc for commons, dfsclient, spark


This would be changed (from how it is now on snapshot repos) of 
org
- hpccsystems
-- HPCC-JAPIs
--- <version dir>
---- pom.xml
-- wsclient
--- <version dir>
---- jars & poms

I think the way I have it set up here (first example) is the correct way.  It insures that the 'top level pom' is in the directory structure above the child modules, and not parallel like in the HPCC-JAPIs directory example.

Thoughts?

drealeed:
The first way (top level hpcc-japis with child wsclient/etc. modules) is how I have our multi-module salt and multi-module hipie projects set up. It always deploys the top level pom as well with no issue.

rpastrana:
What are the side effects of moving this pom up to the org.hpccsystems level? If a group of new hpcc projects decide to live in the org.hpccsystems namespace, and would like to be grouped together like the hpcc-japi projects, but will have different versions, etc than the japi projects, would this preclude them from being grouped as such? Or would we be able to accommodate them within this top level pom? @Michael-Gardner

Michael-Gardner:
Well, we control the namespace under org.hpccsystems.  So I don't think we'll have a problem in the future creating something under the namespace (ex: org.hpccsystems.mynewmodule) that has no dependency on the top level pom at org.hpccsystems.  To my knowledge it should all work fine.  Perhaps @drealeed might know more on the subject though.

rpastrana:
@Michael-Gardner sure, but if other projects under org.hpccsystems need a parent pom for the same reasons as the hpcc-japi projects, where would the parent pom live? Does this top level pom imply owner ship over all sub projects?

rpastrana:
@drealeed does your top level pom act as the parent pom for all the sub projects in your namespace?

drealeed:
yes, it does. Top level pom is com.relx.rba.tardis, and all the submodules live underneath .tardis (.tardis.hpcc, .tardis.visualization etc0

commons-hpcc/pom.xml
```diff
@@ -7,8 +7,8 @@
   <description>Common library for HPCC functionality</description>
 
   <parent>
-    <groupId>org</groupId>
-    <artifactId>hpccsystems</artifactId>
+    <groupId>org.hpccsystems</groupId>
+    <artifactId>HPCC-JAPIs</artifactId>
     <version>7.6.4-SNAPSHOT</version>
```
drealeed: convention--artifact ID should be lower case

Michael-Gardner: Thanks, updated.

rpastrana: @Michael-Gardner as discussed, we should name this artifact japis rather than hpcc-japis since it's part of the org.hpccsystems namespace.

PR #336, opened 2019-11-21
JAPI-236 added go_gold and go_snapshot tagging and branch creation scripts

Michael-Gardner:
Signed-off-by: Michael Gardner <michael.gardner@lexisnexisrisk.com>

## Type of change:
- [x] This change is a breaking change (fix or feature that will cause existing behavior to change).

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/commons-hpcc/pom.xml
+++ b/commons-hpcc/pom.xml
@@ -9,7 +9,7 @@
   <parent>
     <groupId>org.hpccsystems</groupId>
     <artifactId>japi</artifactId>
-    <version>7.6.21-SNAPSHOT</version>
+    <version>7.6.25-0-SNAPSHOT</version>
   </parent>
 
   <build>
--- a/dfsclient/pom.xml
+++ b/dfsclient/pom.xml
@@ -9,7 +9,7 @@
   <parent>
     <groupId>org.hpccsystems</groupId>
     <artifactId>japi</artifactId>
-    <version>7.6.21-SNAPSHOT</version>
+    <version>7.6.25-0-SNAPSHOT</version>
   </parent>
 
   <build>
@@ -43,12 +43,12 @@
     <dependency>
       <groupId>org.hpccsystems</groupId>
       <artifactId>commons-hpcc</artifactId>
-      <version>7.6.21-SNAPSHOT</version>
+      <version>7.6.25-0-SNAPSHOT</version>
     </dependency>
     <dependency>
       <groupId>org.hpccsystems</groupId>
       <artifactId>wsclient</artifactId>
-      <version>7.6.21-SNAPSHOT</version>
+      <version>7.6.25-0-SNAPSHOT</version>
     </dependency>
   </dependencies>
 
--- a/pom.xml
+++ b/pom.xml
@@ -4,7 +4,7 @@
   <modelVersion>4.0.0</modelVersion>
   <groupId>org.hpccsystems</groupId>
   <artifactId>japi</artifactId>
-  <version>7.6.21-SNAPSHOT</version>
+  <version>7.6.25-0-SNAPSHOT</version>
   <packaging>pom</packaging>
   <name>HPCC Systems Java Projects</name>
   <url>https://hpccsystems.com</url>
--- a/wsclient/pom.xml
+++ b/wsclient/pom.xml
@@ -9,7 +9,7 @@
   <parent>
     <groupId>org.hpccsystems</groupId>
     <artifactId>japi</artifactId>
-    <version>7.6.21-SNAPSHOT</version>
+    <version>7.6.25-0-SNAPSHOT</version>
   </parent>
 
   <properties>
@@ -131,7 +131,7 @@
     <dependency>
       <groupId>org.hpccsystems</groupId>
       <artifactId>commons-hpcc</artifactId>
-      <version>7.6.21-SNAPSHOT</version>
+      <version>7.6.25-0-SNAPSHOT</version>
     </dependency>
   </dependencies>
 


Michael-Gardner:
@rpastrana Please review

go_snapshot.sh
```diff
@@ -0,0 +1,64 @@
...
+  fi
+  doit "git checkout -b candidate-$POM_MAJOR.$POM_MINOR.$NEW_POINT"
+  doit "git checkout $GIT_BRANCH"
+  doit "git submodule update --init --recursive"
+  update_version_file SNAPSHOT $((NEW_POINT+1)) $NEW_MINOR
+  doit "git commit -s -a -m \"Split off $POM_MAJOR.$POM_MINOR.$NEW_POINT\""
+  doit "git push $REMOTE"
+  GIT_BRANCH=candidate-$POM_MAJOR.$POM_MINOR.$NEW_POINT
+  doit "git checkout $GIT_BRANCH"
+  doit "git submodule update --init --recursive"
```
rpastrana: not sure this is necessary

go_snapshot.sh
```diff
@@ -0,0 +1,64 @@
...
+  fi
+  if [ "$GIT_BRANCH" != "candidate-$POM_MAJOR.$POM_MINOR.x" ]; then
+    echo "Current branch should be candidate-$POM_MAJOR.$POM_MINOR.x"
+    exit 2
+  fi
+  doit "git checkout -b candidate-$POM_MAJOR.$POM_MINOR.$NEW_POINT"
+  doit "git checkout $GIT_BRANCH"
+  doit "git submodule update --init --recursive"
+  update_version_file SNAPSHOT $((NEW_POINT+1)) $NEW_MINOR
+  doit "git commit -s -a -m \"Split off $POM_MAJOR.$POM_MINOR.$NEW_POINT\""
```
rpastrana: does the message get a sign-off statement

parse_pom.sh
```diff
@@ -0,0 +1,134 @@
...
+if [ ! -f $VERSIONFILE ]; then
+  echo "File $VERSIONFILE not found"
+  exit 2
+fi
+
+GIT_BRANCH=$(git rev-parse --symbolic-full-name --abbrev-ref HEAD)
+
+function parse_pom()
+{
+  POM_PROJECT=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)
```
rpastrana: let's specify this mvn dep on the file header

go_snapshot.sh
rpastrana: judging by Richard's automated commits, there is a sign off:
https://github.com/hpcc-systems/HPCC-Platform/commit/9cedb66e3e60b95263433aefb432cca4f1e42070

parse_pom.sh
```diff
@@ -0,0 +1,139 @@
...
+if [ "$#" -ge 1 ] ; then
+  VERSIONFILE=$1
+  shift 1
+fi
+if [ ! -f $VERSIONFILE ]; then
+  echo "File $VERSIONFILE not found"
+  exit 2
+fi
+
+if ! `which mvn 1>/dev/null 2>&1` ; then
```
Michael-Gardner: Added a comment about needing maven and a check for the dependency.

parse_pom.sh
Michael-Gardner: Added at start of file, and a test for 'which mvn'

go_snapshot.sh
Michael-Gardner: Removed

go_snapshot.sh
```diff
@@ -0,0 +1,64 @@
...
+  else
+    NEW_POINT=$((POM_POINT+1))
+  fi
+  if [ "$GIT_BRANCH" != "candidate-$POM_MAJOR.$POM_MINOR.x" ]; then
+    echo "Current branch should be candidate-$POM_MAJOR.$POM_MINOR.x"
+    exit 2
+  fi
+  doit "git checkout -b candidate-$POM_MAJOR.$POM_MINOR.$NEW_POINT"
+  doit "git checkout $GIT_BRANCH"
+  doit "git submodule update --init --recursive"
```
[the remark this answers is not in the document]
Michael-Gardner: Removed

rpastrana:
@richardkchapman per offline conversation

rpastrana:
@Michael-Gardner what's the status on this pr?

rpastrana:
@Michael-Gardner please confirm these scripts were used successfully, and we're comfortable making them official.

parse_pom_sequence.sh
```diff
@@ -0,0 +1,83 @@
...
+#
+# Common utilities used by tools for automating tagging and release
+#
+
+#We want any failures to be fatal
+# Depends: maven
+
+set -e
+
+if ! `which mvn 1>/dev/null 2>&1` ; then
```
rpastrana: This is valid if we force the user to have mvn on the PATH, and this will prob cover a vast majority of our cases, but it might be worthwhile allowing users to set, or pass in a mvn_home path

parse_pom_sequence.sh
```diff
@@ -0,0 +1,83 @@
...
+  echo "Maven dependency not located"
+  exit 2
+fi
+
+# overwrite versionfile to pom.xml
```
rpastrana: I see the term overwrite, but I'm not sure of the inheritance model in bash, are we actually inheriting from another bash? if so, do we need to declare this fact?

parse_pom_sequence.sh
```diff
@@ -0,0 +1,83 @@
...
+  exit 2
+fi
+
+
+# overwrite parse_cmake to target pom files
+# expect ex: 7.6.2-0-SNAPSHOT, 7.6.2-1 (gold release), etc
+# Major.Minor.Point- Sequence if gold, Maturity if rc
+function parse_cmake()
+{
+  HPCC_PROJECT=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)
```
rpastrana: this mvn command will act on the local pom? It might be worthwhile to log the fully qualified name of the effective pom file.

parse_pom_sequence.sh
```diff
@@ -0,0 +1,83 @@
...
+  HPCC_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
+  HPCC_MAJOR=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $1};')
+  HPCC_MINOR=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $2};')
+  HPCC_POINT=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $3};')
+  HPCC_SEQUENCE=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $4};')
+  HPCC_MATURITY=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $5};')
```
rpastrana: If any of these HPCC_XYZ vars end up as empty strings, we should log and stop the process

parse_pom_sequence.sh
Michael-Gardner: Added in a catch which will error out if we're using the versioning system without a <sequence>

parse_pom_sequence.sh
Michael-Gardner: Added in a catch which will error out if we're using the versioning system without a HPCC_SEQUENCE

parse_pom_sequence.sh
Michael-Gardner: based on Richard's workflow, we'll be in the root directory of the project where the pom is located.

parse_pom_sequence.sh
Michael-Gardner: We're basically just overwriting function definitions from the parse_cmake.sh file that gets sourced earlier by go_rc or go_gold.  The new functions keep the same parameters and output behavior, but are specific to working on a pom.  Only gets sourced IF the pom is present in the local directory we're working from (which is how Richard does his workflow, his command cd's into the local directory and commands from that location)

Michael-Gardner:
In order to make this work seamlessly, I had to overwrite some behavior in the parse_cmake.sh script which gets sourced by HPCC-Platform/cmake_modules/[go_rc.sh, go_gold.sh].  The version.cmake file uses sequences and maturity, which we don't normally use in pom projects.  This also means we'll need to tweak our versioning system slightly like in this PR.  The new scheme will be major.minor.point-sequence[-SNAPSHOT].

`function parse_cmake()
{
  HPCC_PROJECT=$(mvn help:evaluate -Dexpression=project.artifactId -q -DforceStdout)
  HPCC_VERSION=$(mvn help:evaluate -Dexpression=project.version -q -DforceStdout)
  HPCC_MAJOR=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $1};')
  HPCC_MINOR=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $2};')
  HPCC_POINT=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $3};')
  HPCC_SEQUENCE=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $4};')
  HPCC_MATURITY=$(echo $HPCC_VERSION | awk 'BEGIN {FS="[.-]"}; {print $5};')

  if ! `echo $HPCC_SEQUENCE | grep -Eq ^[0-9]+$`; then
    echo "Error: HPCC_SEQUENCE not an Integer"
    echo "-- possibly on branch using old versioning scheme"
    exit 2
  fi

  # translate pom maturity to what go_rc/go_gold understand
  if [ -z "$HPCC_MATURITY" ] ; then
    # branch on gold release
    HPCC_MATURITY=release
  elif [ "$HPCC_MATURITY" == "SNAPSHOT" ] ; then
    if (( "$HPCC_POINT" % 2 != 1 )) ; then
      HPCC_MATURITY=rc
    else
      HPCC_MATURITY=closedown
    fi
  fi

}`

Since we haven't used 'Sequence' in the past, I added a check to ensure that the script finds an integer at the correct location in the version.  If it doesn't, it is likely the branch is using the old versioning scheme.
Maturity is dealt with by making a few assumptions. We have to translate the lack of a maturity or the 'SNAPSHOT' label into something the go_gold and go_rc scripts will understand. Anything that doesn't have a 'maturity' is assumed to be a release, since normally for pom projects we always keep SNAPSHOT present until we tag golds.  If we do find SNAPSHOT and the point mod 2 is odd (ex in candidate-7.6.x we will have a point version at 25), we translate this to 'closedown'.  Otherwise we translate the SNAPSHOT label to 'rc'.

`function set_tag()
{
    local _prefix=$1
    local _maturity=$HPCC_MATURITY
    if [ "$HPCC_MATURITY" == "rc" ] ; then
      _maturity=SNAPSHOT
    fi
    # will keep -release or -SNAPSHOT as maturity for tagging
    HPCC_SHORT_TAG=${HPCC_MAJOR}.${HPCC_MINOR}.${HPCC_POINT}-${HPCC_SEQUENCE}-${_maturity}
    HPCC_LONG_TAG=${_prefix}_$HPCC_SHORT_TAG
}`

The set_tag function has been modified to translate 'rc' back to -SNAPSHOT.  This will result in tags like japi_7.6.24-1-release or japi_7.6.26-1-SNAPSHOT.

`function update_version_file()
{
    # Update the pom.xml file
    local _new_maturity=$1
    local _new_point=$2
    local _new_sequence=$3
    local _new_minor=$4
    if [ -z "$_new_minor" ] ; then
      _new_minor=$HPCC_MINOR
    fi
    if [ "$_new_maturity" == "rc" ]; then
      _new_maturity=-SNAPSHOT
    elif [ "$_new_maturity" == "closedown" ]; then
      _new_maturity=-SNAPSHOT
    else
      # don't set for non-snapshots
      _new_maturity=
    fi
    local _v="${HPCC_MAJOR}.${_new_minor}.${_new_point}-${_new_sequence}${_new_maturity}"
    local mvn_version_update_cmd="mvn versions:set -DnewVersion=$_v"
    if [ -n "$VERBOSE" ] ; then
      echo  "$mvn_version_update_cmd"
    fi
    if [ -z "$DRYRUN" ] ; then 
      eval "$mvn_version_update_cmd"
      find . -name 'pom.xml' | xargs git add
    else
      echo  "Update to $_v"
    fi
}`

The update_version_file function gets overwritten to use maven calls.  If the new_maturity passed in is 'rc' or 'closedown', we translate it to -SNAPSHOT for the final version string, otherwise we strip the maturity for a 'release' which won't have anything appended after the sequence.
The 'mvn versions:set' command will modify the version in our pom and the dependencies it sees if we're in a multi module project like JAPI.  In projects like Spark where it isn't driven by some top level pom that knows about multiple modules and their changing versions, I have a simple workaround that will be included in the PR which will replace the individual 'version' tags for each org.hpccsystems dependency with ${project.version}.

Michael-Gardner:
@rpastrana Looks good for maven central and jenkins

PR #359, opened 2020-02-07
Japi-277 binaryrecordwriter npes

drealeed:
## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Created a unit test that tested writing null values for ecl types STRING, BOOLEAN, INTEGER2, INTEGER8,UNSIGNED8, REAL8, DECIMAL5_2, DATA, UTF8, UNICODE and ENUM. Default values ("" for string, false for boolean, 0 for null numerics and enum) were correctly written.

Also reran all other ReadWriteDFUTest use cases.

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
@@ -22,6 +22,7 @@ import org.hpccsystems.commons.ecl.FieldType;
 import org.hpccsystems.commons.ecl.HpccSrcType;
 
 import java.math.BigDecimal;
+import java.math.BigInteger;
 import java.nio.ByteBuffer;
 import java.nio.ByteOrder;
 import java.io.OutputStream;
@@ -257,7 +258,7 @@ public class BinaryRecordWriter implements IRecordWriter
             }
             case BOOLEAN:
             {
-                Boolean value = (Boolean) fieldValue;
+                Boolean value = fieldValue==null?Boolean.valueOf(false):(Boolean) fieldValue;
                 byte byteValue = 0;
                 if (value)
                 {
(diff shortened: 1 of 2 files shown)


dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -269,14 +270,21 @@ private void writeField(FieldDef fd, Object fieldValue) throws Exception
             case INTEGER:
             {
                 Long value = null;
-                if (fieldValue instanceof Long)
+                if (fieldValue==null) {
```
rpastrana: format - new line

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -269,14 +270,21 @@ private void writeField(FieldDef fd, Object fieldValue) throws Exception
...
                 {
                     value = (Long) fieldValue;
                 }
                 else if (fieldValue instanceof Integer)
                 {
                     value = ((Integer) fieldValue).longValue();
                 }
+                else if (fieldValue instanceof BigInteger)
+                {
+                    value = ((BigInteger) fieldValue).longValue();
```
rpastrana: can fieldValue == null?, default to 0?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -718,7 +730,7 @@ private void writeDecimal(FieldDef fd, BigDecimal decimalValue)
 
         // Strip leading zeros. We only want sig digits
         int numLeadingZeros = 0;
-        while (decimalStr.charAt(numLeadingZeros) == '0')
+        while (decimalStr.length()>numLeadingZeros && decimalStr.charAt(numLeadingZeros) == '0')
```
rpastrana: do we need to ensure decimalValue != null early on?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
drealeed: that's trapped by the if fieldvalue==null case above it, it'll never be null at this point

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
drealeed: This method's called only from writeField, and the line before it's callled makes sure it's not null:
BigDecimal value = fieldValue==null?BigDecimal.valueOf(0):(BigDecimal) fieldValue;
writeDecimal(fd, value);

PR #360, opened 2020-02-07
Japi-278 handle npes in RowServiceInputStream

drealeed:
## Type of change:

## Checklist:
- [x] My change requires a change to the documentation.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Created a unit test; also reran DFUReadWriteTest.

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
@@ -579,7 +579,10 @@ public class RowServiceInputStream extends InputStream
     {
         try
         {
-            int cursorLen = dis==null?0:dis.readInt();
+            if (dis==null) {
+                return;
+            }
+            int cursorLen = dis.readInt();
             if (cursorLen == 0)
             {
                 close();
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
@@ -579,7 +579,7 @@ public class RowServiceInputStream extends InputStream
     {
         try
         {
-            int cursorLen = dis.readInt();
+            int cursorLen = dis==null?0:dis.readInt();
             if (cursorLen == 0)
             {
                 close();
@@ -708,7 +708,9 @@ public class RowServiceInputStream extends InputStream
             }
 
             this.dos.close();
-            this.dis.close();
+            if (this.dis != null) {
+                this.dis.close();
+            }
             this.sock.close();
             this.dos = null;
             this.dis = null;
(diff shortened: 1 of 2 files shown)


dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -708,7 +708,10 @@ public void close() throws IOException
             }
 
             this.dos.close();
-            this.dis.close();
+            if (this.dis != null)
```
rpastrana: Are the output stream and sock guaranteed to be non-null here?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -61,6 +62,17 @@ public void setup() throws Exception
         super.setup();
     }
 
+    
+    @Test
+    public void readBadlyDistributedFileTest() throws Exception
+    {
+        //this file only has data on two nodes
+        HPCCFile file = new HPCCFile("~test::alldatatypes", connString , hpccUser, hpccPass);
```
rpastrana: is "test::alldatatypes" inherently skewed, or should we ensure skewed distribution?

drealeed: yes, it's inherently skewed (all data is on two nodes)

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
drealeed: I believe so, as the output stream is created by us. The input stream can be null if there are no records on the node. I'm fine waiting until @jpmcmu  is back to verify this.

jpmcmu:
@drealeed one change request, otherwise it looks good.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -579,7 +579,7 @@ private void finishFetch()
     {
         try
         {
-            int cursorLen = dis.readInt();
+            int cursorLen = dis==null?0:dis.readInt();
```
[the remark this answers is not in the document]
drealeed: what about closing the output stream and socket? Is it safe to assume that if dis is null they'll be null too?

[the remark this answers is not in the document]
jpmcmu: @drealeed Yep, if dis is null then close() has been called and the output stream and socket should be null as well. I need to cleanup that flow, but checking if dis is null and then bailing should fix your issue.

jpmcmu:
@drealeed @rpastrana looks good

PR #370, opened 2020-03-20
JAPI-285 DFSClient: HPCCFile setFilter does not work if partitions ha…

jpmcmu:
…ve already been created

- Corrected setFilter to modify exisiting dataparts
- Added setFilter method to DataPartition

Signed-off-by: James McMullan <James.McMullan@lexisnexis.com>

## Type of change:

## Checklist:
  - [X] I have applied the Eclipse code-format template provided.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/HPCCFile.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/HPCCFile.java
@@ -46,6 +46,8 @@ public class HPCCFile implements Serializable
     private static final Logger  log                           = LogManager.getLogger(HPCCFile.class);
 
     private DataPartition[]      dataParts;
+    private long                 dataPartsCreationTimeMS       = -1;
+
     private FieldDef             recordDefinition;
     private FieldDef             projectedRecordDefinition;
     private boolean              isIndex;
@@ -191,6 +193,10 @@ public class HPCCFile implements Serializable
     public HPCCFile setFileAccessExpirySecs(int fileAccessExpirySecs)
     {
         this.fileAccessExpirySecs = fileAccessExpirySecs;
+
+        // Force the data parts to be recreated
+        this.dataParts = null;
+
         return this;
     }
 
@@ -214,6 +220,10 @@ public class HPCCFile implements Serializable
     public HPCCFile setTargetfilecluster(String targetfilecluster)
     {
         this.targetfilecluster = targetfilecluster;
+
+        // Force the data parts to be recreated
+        this.dataParts = null;
+
         return this;
     }
 
@@ -237,6 +247,10 @@ public class HPCCFile implements Serializable
     public HPCCFile setClusterRemapInfo(RemapInfo remapinfo)
     {
         this.clusterRemapInfo = remapinfo;
+
+        // Force the data parts to be recreated
+        this.dataParts = null;
+
         return this;
     }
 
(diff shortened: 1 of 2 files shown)


jpmcmu:
@rpastrana please review

dfsclient/src/main/java/org/hpccsystems/dfs/client/HPCCFile.java
```diff
@@ -262,6 +262,15 @@ public FileFilter getFilter()
     public HPCCFile setFilter(String filterexpression) throws Exception
     {
         this.filter = new FileFilter(filterexpression);
+
```
rpastrana: do the other setters setProjectList, setClusterRemapInfo, setTargetfilecluster, setFileAccessExpirySecs need to update private members?

jpmcmu: Nope, this was the only setter that affected the DataPartition

jpmcmu: We do have some HpccFile filter unit tests on the Spark side, but they didn't hit the ordering issue that the DataSource exposed. It is probably a good idea to move those over to the DFSClient unit tests and expand them a bit. I will create a JIRA

rpastrana: Although I'd like to get rid of the "clusterremapper", it seems to affect the parts:
ClusterRemapper clusterremapper = ClusterRemapper.makeMapper(clusterRemapInfo, fileinfoforread);
this.dataParts = DataPartition.createPartitions(fileinfoforread.getFileParts(), clusterremapper,0, filter, fileinfoforread.getFileAccessInfoBlob(), fileType);

Perhaps the remapper cannot be set once the fileparts have already been created, same thing for the setTargetfilecluster, and setFileAccessExpirySecs

jpmcmu: You are right clusterremapper would affect things, I missed that when I first looked at it. 
Cluster remapper complicates things, because it is destructive in that we won't have the original IPs after it does its job. I will look into the viability of not generating the data partitions until they are explicitly asked for. That would still potentially leave us with an issue for clusterremapper changes, I guess it can default to recreating the data partitions or throwing an error.

rpastrana: @jpmcmu those setters should error out if the parts have already been created

dfsclient/src/main/java/org/hpccsystems/dfs/client/HPCCFile.java
```diff
@@ -283,6 +306,15 @@ public String getFileName()
      */
     private void createDataParts() throws HpccFileException
     {
+        long fileAccessExpiryMS = fileAccessExpirySecs * 1000;
+        long dataPartsAgeMS = System.currentTimeMillis() - dataPartsCreationTimeMS;
+        if (dataParts != null && dataPartsAgeMS < fileAccessExpiryMS)
+        {
+            return;
+        }
```
rpastrana: if datapars !=null, but dataPartsAgeMS  >= fileAccessExpiryMS, let's output a trace statement explaining why we're refreshing the hpccfile

jpmcmu: @rpastrana updated to log when refresh due to expiration

PR #383, opened 2020-04-24
JAPI-297 Generated wrappers improvement

rpastrana:
- Updates generator logic to implemente improvements
- Wrapped fields initialized to null
- Wrapped fields only instantiated if raw != null
- Several format improvements

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


rpastrana:
@drealeed please review. Although I had just recently mentioned to you the generated wrappers shouldn't change often, I'm updating most of them here. I found we have been wastefully initializing wrapped fields which in some cases are rather large and several levels nested. 
Please focus on the new logic in Axis2ADBStubWrapperMaker.java, look over some of wrappers of most use and interest to you, and take a look at the change in FileSprayClientTest which might illustrate the type of change you might need to make on your side. Thanks.

wsclient/src/main/java/org/hpccsystems/ws/client/wrappers/gen/wssql/ArrayOfECLTimerWrapper.java
```diff
@@ -72,8 +72,9 @@ public String toString()
 			for ( int i = 0; i < this.local_eCLTimer.size(); i++)
 			{
```
drealeed, line 73: now that these values are null rather than initialized, won't loops like this potentially throw an NPE?

drealeed:
Rodrigo, I'll need to run our hpcc unit tests against this to see if any new NPEs occur. For example, a lot of our code assumes that the exceptions collection returned by a result wrapper will always be empty if there's no exceptions rather than null.

rpastrana:
> 
> 
> Rodrigo, I'll need to run our hpcc unit tests against this to see if any new NPEs occur. For example, a lot of our code assumes that the exceptions collection returned by a result wrapper will always be empty if there's no exceptions rather than null.

Absolutely, you prob have code similar to the FileSprayClientTest code I changed in this PR...
The exceptions wrapper isn't very deep, I don't mind initializing that one if it minimizes the potential for NPEs, your thoughts?

wsclient/src/main/java/org/hpccsystems/ws/client/wrappers/gen/wssql/ArrayOfECLTimerWrapper.java
rpastrana: line 69 prevents NPE, but yes let's keep an eye out for any potential reference to uninitialized vars.

rpastrana:
@drealeed did you encounter any issues w/ this change?

drealeed:
all hpcc unit tests pass

PR #399, opened 2020-07-21
JAPI-318 : provide way to define connect timeout when writing files.

drealeed:
* Create additional constructor for HPCCRemoteFileWriter that takes a connect timeout value
* update current constructor for HPCCRemoteFileWriter to call the new constructor, passing RowServiceOutputWriter.DEFAULT_CONNECT_TIMEOUT_MILLIS
* Create additional constructor for RowServiceOutputWriter that takes a connect timeout value
* update current (non-deprecated) constructor for RowServiceOutputWriter to call the new constructor, passing RowServiceOutputWriter.DEFAULT_CONNECT_TIMEOUT_MILLIS
* update DFSReadWriteTest so that it can run passing a connect timeout value defined as a system property. If no connect timeout is defined, the test runs as it used do with default timeouts specified.

## Type of change:

## Checklist:
  - [] I have applied the Eclipse code-format template provided.
- [X] My change requires a change to the documentation.
  - [X] Any new interfaces or exported functions are appropriately commented.
  - [X] I have performed system test and covered possible regressions and side effects.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

## Testing:

Tested the change by running DFUReadWriteTest with and without the connecttimeoutms system property specified.

 Also, compiled the jar with changes and tested it from the tardis hpcc module. Confirmed that these changes resolved the issues tardis was experiencing with write connection timeouts.

### DIFF
--- a/dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
+++ b/dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
@@ -61,10 +61,10 @@ public class DFSReadWriteTest extends BaseRemoteTest
     public void readBadlyDistributedFileTest() throws Exception
     {
         //this file only has data on two nodes
-        HPCCFile file = new HPCCFile("~benchmark::all_types::200kb", connString , hpccUser, hpccPass);
+        HPCCFile file = new HPCCFile(datasets[1], connString , hpccUser, hpccPass);
         file.setProjectList("");
-        List<HPCCRecord> records = readFile(file);
-        assertEquals("Not all records loaded",records.size(),5600);
+        List<HPCCRecord> records = readFile(file, connTO);
+        assertEquals("Not all records loaded",expectedCounts[1], records.size());
     }
 
     @Test
@@ -75,7 +75,7 @@ public class DFSReadWriteTest extends BaseRemoteTest
             HPCCFile file = new HPCCFile(datasets[i], connString , hpccUser, hpccPass);
             file.setProjectList("");
 
-            List<HPCCRecord> records = readFile(file);
+            List<HPCCRecord> records = readFile(file, connTO);
             if (records.size() != expectedCounts[i])
             {
                 Assert.fail("Record count mismatch for dataset: " + datasets[i] + " got: " + records.size() + " expected: " + expectedCounts[i]);
@@ -83,12 +83,12 @@ public class DFSReadWriteTest extends BaseRemoteTest
 
             // Write the dataset back
             String copyFileName = datasets[i] + "-copy13";
-            writeFile(records, copyFileName, file.getProjectedRecordDefinition());
+            writeFile(records, copyFileName, file.getProjectedRecordDefinition(),connTO);
 
             // Read and compare to original dataset
             file = new HPCCFile(copyFileName, connString , hpccUser, hpccPass);
 
-            List<HPCCRecord> copiedRecords = readFile(file);
+            List<HPCCRecord> copiedRecords = readFile(file, connTO);
             if (copiedRecords.equals(records) == false)
             {
                 Assert.fail("Written dataset does not match original dataset: " + copyFileName);
@@ -105,7 +105,7 @@ public class DFSReadWriteTest extends BaseRemoteTest
 
             FieldDef recdef=file.getRecordDefinition();
             file.setProjectList(StringUtils.join(projectedfields, ","));
-            List<HPCCRecord> recs=readFile(file);
+            List<HPCCRecord> recs=readFile(file, connTO);
             if (recs.get(0).getNumFields() != file.getRecordDefinition().getNumDefs()-1) 
             {
                 fail("recs did not project correctly");
@@ -120,7 +120,7 @@ public class DFSReadWriteTest extends BaseRemoteTest
         String fname = datasets[1]; 
         HPCCFile file = new HPCCFile(fname, connString, hpccUser, hpccPass);
         file.setProjectList("");
-        List<HPCCRecord> records = readFile(file);
+        List<HPCCRecord> records = readFile(file, connTO);
         assertEquals( "Record count mismatch for dataset:" + fname + ". got: " + records.size() + " expected:" + expectedCounts[1],expectedCounts[1],records.size());
 
         HPCCRecord first = records.get(0);
@@ -131,11 +131,12 @@ public class DFSReadWriteTest extends BaseRemoteTest
 
         // Write the dataset back
         String copyFileName = fname + "_copy";
-        writeFile(records, copyFileName, file.getProjectedRecordDefinition());
+        writeFile(records, copyFileName, file.getProjectedRecordDefinition(),connTO);
 
         file = new HPCCFile(copyFileName, connString, hpccUser, hpccPass);
-        records = readFile(file);
-        assertEquals( "Record count mismatch for dataset:" + copyFileName + ". got: " + records.size() + " expected:" + 5600,5600,records.size());
+        records = readFile(file, connTO);
+        assertEquals( "Record count mismatch for dataset:" + copyFileName + ". got: " + records.size() +
+                " expected:" + expectedCounts[1],expectedCounts[1],records.size());
 
         HPCCRecord rec0 = records.get(0);
         for (int i = 0; i < rec0.getNumFields(); i++)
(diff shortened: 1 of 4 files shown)


dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -342,8 +345,15 @@ private void writeFile(List<HPCCRecord> records, String fileName, FieldDef recor
                 }
 
                 HPCCRecordAccessor recordAccessor = new HPCCRecordAccessor(recordDef);
-                HPCCRemoteFileWriter<HPCCRecord> fileWriter = new HPCCRemoteFileWriter<HPCCRecord>(hpccPartitions[partitionIndex], recordDef,
-                        recordAccessor, CompressionAlgorithm.NONE);
+                HPCCRemoteFileWriter<HPCCRecord> fileWriter = null;
+                if (connectTimeoutMs != null) {
+                    fileWriter=new HPCCRemoteFileWriter<HPCCRecord>(hpccPartitions[partitionIndex], recordDef,
+                            recordAccessor, CompressionAlgorithm.NONE,connectTimeoutMs);
+                    Thread.sleep(2000);
```
rpastrana, line 352: what are we accomplishing with this sleep?
Also, minor but formatting needed.

wsclient/src/main/java/org/hpccsystems/ws/client/platform/test/BaseRemoteTest.java
```diff
@@ -42,7 +42,7 @@ HPCC SYSTEMS software Copyright (C) 2019 HPCC Systems®.
 
     protected final static String hpccUser = System.getProperty("hpccuser", defaultUserName);
     protected final static String hpccPass = System.getProperty("hpccpass", "");
-    protected final static String connTO = System.getProperty("connecttimeoutmillis");
+    protected final static Integer connTO = System.getProperty("connecttimeoutmillis")==null?null:Integer.valueOf(System.getProperty("connecttimeoutmillis"));
```
rpastrana, line 45: are all references to connTO in the project null safe?
is the ternary expression needed because Integer.valueOf(null) != null?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -315,7 +318,7 @@ private void writeFile(List<HPCCRecord> records, String fileName, FieldDef recor
 
             String clusterName = this.thorcluster;
             System.out.println("Create Start");
-            DFUCreateFileWrapper createResult = dfuClient.createFile(fileName, clusterName, eclRecordDefn, 300, false, DFUFileTypeWrapper.Flat, "");
+            DFUCreateFileWrapper createResult = dfuClient.createFile(fileName, clusterName, eclRecordDefn, connTO, false, DFUFileTypeWrapper.Flat, "");
```
rpastrana, line 321: Not sure if important, but this test's connection timeout is different now

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -234,25 +235,27 @@ public void filteredTest() throws Exception
...
-    public List<HPCCRecord> readFile(HPCCFile file) throws Exception
+    public List<HPCCRecord> readFile(HPCCFile file, Integer connectTimeoutMillis) throws Exception
     {
         if (file == null)
         {
             Assert.fail("HPCCFile construction failed.");
         }
-
+        if (connectTimeoutMillis != null) {
+            file.setFileAccessExpirySecs(connectTimeoutMillis/1000);
```
rpastrana, line 257: typically the connect timeout would be considerably shorter than the fileaccessexpiry, i'm not sure what we get from combining the two values. Albeit, we prob do need a junit test w/ very short/log expiry values...

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -55,16 +55,16 @@
 public class DFSReadWriteTest extends BaseRemoteTest
 {
     private static final String[] datasets       = { "~benchmark::integer::20kb", "~benchmark::all_types::200kb"};
-    private static final int[]    expectedCounts = { 1250, 5600 };
+    private static final int[]    expectedCounts = { 1250, 1558 };
```
rpastrana, line 58: was the 5600 record count invalid? @jpmcmu

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
drealeed: the default timeout ms is 1000. If an override is specified, sleep longer than the default timeout to verify the overridden timeout is actually working.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
drealeed: Fixed this so that if connTO is null the original 300 is used.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
drealeed: yes, the current file has 1558 recs, I checked on the cluster to confirm.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
drealeed: I was getting a file access expiry error (connecting over vpn) until I added this to the test.

wsclient/src/main/java/org/hpccsystems/ws/client/platform/test/BaseRemoteTest.java
drealeed: Integer.valueOf(null) throws an error. Yes, the code handles connTO being null. I found one place it wasn't and fixed it and pushed the change to this PR.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
drealeed: Fixed formatting

drealeed:
@rpastrana ,all comments and questions fixed/answered.

jpmcmu:
@drealeed @rpastrana beyond record count looks good to me. We are generating that dataset in our Jenkins test environment. So the record count is probably different than what is on the 160.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
rpastrana: @drealeed the datasets and expectcounts should really be passed in dynamically, but for now our official build environment's file has 5600 recs. 
let's revert, and we'll provide you the ECL we execute to generate the larger dataset

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -55,16 +55,16 @@
...
     @Test
     public void readBadlyDistributedFileTest() throws Exception
     {
         //this file only has data on two nodes
         HPCCFile file = new HPCCFile("~benchmark::all_types::200kb", connString , hpccUser, hpccPass);
         file.setProjectList("");
-        List<HPCCRecord> records = readFile(file);
-        assertEquals("Not all records loaded",records.size(),5600);
+        List<HPCCRecord> records = readFile(file, connTO);
+        assertEquals("Not all records loaded",expectedCounts[1], records.size());
```
rpastrana, line 67: Good catch here, can we also update line 64 to reference datasets[1]?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
rpastrana: fair enough, but my issue has to do w/ the reuse of the connTO, we need a dfsclient test specific fileAccessExpiry variable

private final static Integer fileAccessExpiry = System.getProperty("fileAccessExpirySecs")==null?null:Integer.valueOf(System.getProperty("fileAccessExpirySecs"));

drealeed:
updated file reference to use datasets; reset the file count to 5600; added some logic around the thread.sleep to clarify what it's doing,

Once approved I can squash all these (if I do it now comment pointers will be lost), or you can squash while pulling.

rpastrana:
Thanks @drealeed please go ahead and sqaush.

drealeed:
@rpastrana squashed

PR #404, opened 2020-08-13
JAPI-323 BinaryRecordWriter does not handle BigDecimal precision conv…

jpmcmu:
…ersion correctly

- Corrected writeDecimal to handle precision conversion correctly
- Optimized writeDecimal function
- Combined writeDecimal and writeUnsignedDecimal into a single function
- Updated default write socket timeout
- Added in missing exception case

Signed-off-by: James McMullan <James.McMullan@lexisnexis.com>

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
@@ -141,7 +141,14 @@ public class BinaryRecordWriter implements IRecordWriter
                 case STRING:
                 case CHAR:
                 case VAR_STRING:
-                    writeField(fd, fieldValue);
+                    try
+                    {
+                        writeField(fd, fieldValue);
+                    }
+                    catch (Exception e)
+                    {
+                        throw new Exception("Error while writing field: " + fd.getFieldName() + " of type: " + fd.getFieldType() + ": ", e);
+                    }
                     break;
                 case RECORD:
                     writeRecord(recordAccessor.getChildRecordAccessor(i), fieldValue);
@@ -740,7 +747,14 @@ public class BinaryRecordWriter implements IRecordWriter
         {
             for (Object value : list)
             {
-                this.writeField(fd.getDef(0), value);
+                try
+                {
+                    this.writeField(fd.getDef(0), value);
+                }
+                catch (Exception e)
+                {
+                    throw new Exception("Error while writing field: " + fd.getFieldName() + " of type: " + fd.getFieldType() + ": ", e);
+                }
             }
         }
         else
@@ -751,7 +765,7 @@ public class BinaryRecordWriter implements IRecordWriter
             }
         }
     }
-    
+
     /**
      * Write decimal.
      *
@@ -797,7 +811,8 @@ public class BinaryRecordWriter implements IRecordWriter
         BigInteger unscaledInt = decimalValue.scaleByPowerOfTen(desiredScale).toBigIntegerExact();
 
         int signOffset = 4;
-        if (fd.isUnsigned()) {
+        if (fd.isUnsigned())
+        {
             signOffset = 0;
         }
 
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
@@ -332,15 +332,8 @@ public class BinaryRecordWriter implements IRecordWriter
             }
             case DECIMAL:
             {
-                BigDecimal value = fieldValue==null?BigDecimal.valueOf(0):(BigDecimal) fieldValue;
-                if (fd.isUnsigned())
-                {
-                    writeUnsignedDecimal(fd, value);
-                }
-                else
-                {
-                    writeDecimal(fd, value);
-                }
+                BigDecimal value = fieldValue==null ? BigDecimal.valueOf(0) : (BigDecimal) fieldValue;
+                writeDecimal(fd, value);
                 break;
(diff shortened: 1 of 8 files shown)


jpmcmu:
@rpastrana please review

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -768,131 +762,67 @@ private void writeList(FieldDef fd, IRecordAccessor childRecordAccessor, List<Ob
...
+        // Correctly truncating decimal sig digits and dropping int digits larger than precision - scale
+        BigInteger unscaledInt = decimalValue.scaleByPowerOfTen(desiredScale).toBigIntegerExact();
 
-        // Strip leading zeros. We only want sig digits
-        int numLeadingZeros = 0;
-        while (decimalStr.length()>numLeadingZeros && decimalStr.charAt(numLeadingZeros) == '0')
-        {
-            numLeadingZeros++;
+        int signOffset = 4;
+        if (fd.isUnsigned()) {
```
rpastrana, line 800: format

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSRecordBuilderTest.java
```diff
@@ -67,4 +67,5 @@ public void hpccRecordBuilderTest() throws Exception
             Assert.fail("Test failed with error: " + e.getMessage());
         }
     }
+
```
rpastrana, line 70: remove

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -32,7 +32,7 @@
 public class RowServiceOutputStream extends OutputStream
 {
     private static final Logger  log                           = LogManager.getLogger(RowServiceOutputStream.class);
-    public static final int           DEFAULT_CONNECT_TIMEOUT_MILIS = 1000;                                              // 1 second connection timeout
+    public static final int      DEFAULT_CONNECT_TIMEOUT_MILIS = 5000; // 5 second connection timeout
```
rpastrana, line 35: Just curious how we arrived at this default value. Default timeout values are always difficult to get right and there always has to be compromises. In this case, how often would we encounter this timeout on a single request? once per req, or perhaps once per req?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -378,6 +371,7 @@ else if (fd.getDataLen() == 8)
                     c = (byte) value.charAt(0);
                 }
                 this.buffer.put(c);
+                break;
```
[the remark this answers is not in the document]
jpmcmu: Yeah, I don't think that was a case we were exercising anywhere. I will add that to a unit test.

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -768,131 +762,67 @@ private void writeList(FieldDef fd, IRecordAccessor childRecordAccessor, List<Ob
...
-
-        for (int i = 0; i < decimalStr.length(); i++)
+        boolean isNegative = decimalValue.signum() == -1;
+        if (fd.isUnsigned())
         {
-            if (decimalStr.charAt(i) == '.')
+            if (isNegative)
             {
-                continue;
+                decimalValue = BigDecimal.ZERO;
```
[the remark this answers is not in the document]
jpmcmu: In this case the DECIMAL is unsigned so I am clamping negative values to zero

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordWriter.java
```diff
@@ -768,131 +762,67 @@ private void writeList(FieldDef fd, IRecordAccessor childRecordAccessor, List<Ob
...
-        int scale = fd.getScale();
         int dataLen = (int) fd.getDataLen();
-
-        int mostSigDigit = precision - scale;
-        boolean isNegative = decimalValue.signum() == -1;
-        if (isNegative)
+        for (int i = 0; i < dataLen; i++)
         {
-            decimalValue = decimalValue.negate();
+            this.scratchBuffer[i] = 0;
```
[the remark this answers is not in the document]
jpmcmu: Yep, scratch buffer is larger than the max supported dataLen for a decimal, so that should be fine. Zeroing out the scratch buffer should only be required for decimal to do how I am constructing the values. Other types are setting bytes explicitly instead of ORing bits into the buffer

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: Good question. I don't know if this is the best timeout, but I set it to match what we had RowServiceInputStream. When I am on VPN testing there are times where the original 1 sec timeout happened a lot, so I thought it would be good to change that default. 5s seems to work better for the VPN case, but it can be set by the user after that PR that Drea submitted.

drealeed: Yes, when I'm using this class I almost always override it with some constant or other, so I don't think it's a big issue.

PR #408, opened 2020-09-14
JAPI-326 Summarize read operation messages

rpastrana:
- Provide stream operation message class
- Store messages in hashmap, keyed on message type, message and target column
- Report summary of messages upon completion of operation

Signed-off-by: Rodrigo Pastrana <Rodrigo.Pastrana@lexisnexisrisk.com>

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
@@ -17,7 +17,6 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.math.MathContext;
 import java.nio.charset.Charset;
-import java.nio.ByteBuffer;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -29,9 +28,6 @@ import org.hpccsystems.commons.ecl.FieldDef;
 import org.hpccsystems.commons.ecl.HpccSrcType;
 import org.hpccsystems.commons.errors.HpccFileException;
 import org.hpccsystems.commons.errors.UnparsableContentException;
-
-import org.hpccsystems.dfs.client.IRecordBuilder;
-import org.hpccsystems.dfs.client.IRecordReader;
 
 class CountingInputStream extends InputStream
 {
@@ -129,7 +125,7 @@ public class BinaryRecordReader implements IRecordReader
     private CountingInputStream  inputStream;
     private FieldDef             rootRecordDefinition;
     protected boolean            defaultLE;
-    
+
     private byte[]               scratchBuffer = new byte[BUFFER_GROW_SIZE];
 
     private static final Charset sbcSet              = Charset.forName("ISO-8859-1");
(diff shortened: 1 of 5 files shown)


rpastrana:
@jpmcmu please take a look

jpmcmu:
@rpastrana looks like we are missing some files

rpastrana:
@jpmcmu I was only missing the most important file... please take a look now.

dfsclient/src/main/java/org/hpccsystems/dfs/client/StreamOperationMessages.java
```diff
@@ -0,0 +1,104 @@
...
+    {
+        Error,
+        Warning,
+        Info
+    };
+
+    private HashMap<String, Integer> map = new HashMap<>();
+    private int totalMessageCount = 0;
+
+    private static String createKey(String message, MessageType type, int code, String column, String value)
```
jpmcmu, line 35: Should we use an int hash code from String.hashCode() here as the key instead of the String?

dfsclient/src/main/java/org/hpccsystems/dfs/client/StreamOperationMessages.java
```diff
@@ -0,0 +1,104 @@
...
+    {
+        addMessage(message, type, code, null);
+    }
+
+    public void addMessage(String message, MessageType type, int code, String column)
+    {
+        addMessage(message, type, code, column, null);
+    }
+
+    public void addMessage(String message, MessageType type, int code, String column, String value)
```
jpmcmu, line 87: I am slightly concerned about the performance implications of creating a hashed key for each field on something like the uint8 overflow warning; we would probably be calling addMessage() every other field. 

Maybe we can allow users to pre-calculate keys, and create an addMessage() that takes a key?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -312,8 +325,7 @@ private Object parseFlatField(FieldDef fd, boolean isLittleEndian) throws Unpars
                     intValue = getUnsigned((int) fd.getDataLen(), fd.getSourceType() == HpccSrcType.LITTLE_ENDIAN);
                     if (intValue < 0)
                     {
-                        log.warn(
-                                "Detected possible unsigned value overflow in field '" + fd.getFieldName() + "'. Ensure proper value interpretation");
+                        messages.addMessage("Detected possible unsigned value overflow, ensure proper value interpretation",  MessageType.Warning, fd.getFieldName());
```
jpmcmu, line 328: Referencing the comment below, would it be possible to pre-calculate this error key for each field so that we aren't incurring the createKey() cost for each messsage?

jpmcmu (commented):
@rpastrana Couple questions regarding overhead

dfsclient/src/main/java/org/hpccsystems/dfs/client/StreamOperationMessages.java
rpastrana: I agree. I'm not sure if it's the same as your proposal, but I did consider pre-calculated messages based on an error code. The keys would need to be unique per code +field

jpmcmu:
@rpastrana Looks good

PR #444, opened 2020-11-20
JAPI-361 Ability to "restart" a file read

jpmcmu:
## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
@@ -125,6 +125,7 @@ public class BinaryRecordReader implements IRecordReader
     private CountingInputStream  inputStream;
     private FieldDef             rootRecordDefinition;
     protected boolean            defaultLE;
+    private long                 streamPosAfterLastRecord = 0;
 
     private byte[]               scratchBuffer = new byte[BUFFER_GROW_SIZE];
 
@@ -160,6 +161,16 @@ public class BinaryRecordReader implements IRecordReader
         return messages.getTotalMessageCount();
     }
 
+    public long getStreamPosAfterLastRecord()
+    {
+        return this.streamPosAfterLastRecord;
+    }
+
+    public BinaryRecordReader(InputStream is) throws Exception
+    {
+        this(is,0);
+    }
+
     /**
      * A Binary record reader.
      *
@@ -168,9 +179,15 @@ public class BinaryRecordReader implements IRecordReader
      * @throws Exception
      *             the exception
      */
-    public BinaryRecordReader(InputStream is) throws Exception
+    public BinaryRecordReader(InputStream is, long streamPos) throws Exception
     {
+        if (streamPos < 0)
+        {
+            throw new Exception("BinaryRecordReader: invalid initial streamPos provided: " + streamPos);
+        }
+
         this.inputStream = new CountingInputStream(is);
+        this.inputStream.streamPos = streamPos;
         this.defaultLE = true;
 
         if (this.inputStream.markSupported() == false)
@@ -271,6 +288,7 @@ public class BinaryRecordReader implements IRecordReader
             throw new HpccFileException("Failed to parse next record: " + e.getMessage(), e);
         }
 
+        this.streamPosAfterLastRecord = this.inputStream.getStreamPosition();
         return record;
     }
 
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
@@ -14,7 +14,6 @@ package org.hpccsystems.dfs.client;
 
 import org.hpccsystems.commons.ecl.FieldDef;
 import org.hpccsystems.commons.errors.HpccFileException;
-
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 
@@ -38,6 +37,13 @@ public class HpccRemoteFileReader<T> implements Iterator<T>
     public static final int    NO_RECORD_LIMIT                  = -1;
     public static final int    DEFAULT_READ_SIZE_OPTION         = -1;
     public static final int    DEFAULT_CONNECT_TIMEOUT_OPTION   = -1;
+
+    public static class FileReadResumeInfo
+    {
+        public long inputStreamPos = 0;
+        public byte[] tokenBin = null;
+        public long recordReaderStreamPos = 0;
+    };
 
     /**
      * Instantiates a new hpcc remote file reader.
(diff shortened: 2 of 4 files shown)


jpmcmu:
@drealeed I still need to do some testing and a unit case, but I thought I would create this in case you wanted to give it a shot.

drealeed:
This is fantastic.  I hooked it into the tardis FileProxy multi-threaded hpcc reader and it's working. (https://github.com/hpcc-systems/SALT/compare/master...drealeed:restart_hpcc_read?expand=1)

One thing that would be useful is some flag in the FileReadResumeInfo indicating that the file read is complete, so I wouldn't even have to create an HPCCRemoteFileReader for that partition.

One issue I did encounter was that when something bombed, the record most recently retrieved from hpcc wasn't retrieved in the refetch, although in almost all cases I'd want to re-retrieve that record. Would it be possible to also include a "last but one" buffer location in the FileReadResumeInfo and/or some other flag indicating that the last record processed last time should be reread?

I'm really excited to start using this.

drealeed:
I added the following to HPCCRemoteFileReader, as per your email, and was able to use these methods to implement transactional record reading in Tardis.

`````

    public long getStreamPosAfterLastRecord() {
        return this.binaryRecordReader.getStreamPosAfterLastRecord();
    }

    public FileReadResumeInfo getFileReadResumeInfo(Long recordPointer)
    {
        FileReadResumeInfo resumeInfo = new FileReadResumeInfo();
        resumeInfo.recordReaderStreamPos = recordPointer;

        RowServiceInputStream.RestartInformation isRestartInfo = this.inputStream.getRestartInformationForStreamPos(resumeInfo.recordReaderStreamPos);
        resumeInfo.inputStreamPos = isRestartInfo.streamPos;
        resumeInfo.tokenBin = isRestartInfo.tokenBin;

        return resumeInfo;
    }

    public FileReadResumeInfo getFileReadResumeInfo()
    {
        return getFileReadResumeInfo(this.binaryRecordReader.getStreamPosAfterLastRecord());
    }
`````

File: dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java [hunk 1]
```diff
@@ -168,9 +179,10 @@ public int getStreamMessageCount()
      * @throws Exception
      *             the exception
      */
-    public BinaryRecordReader(InputStream is) throws Exception
+    public BinaryRecordReader(InputStream is, long streamPos) throws Exception
     {
         this.inputStream = new CountingInputStream(is);
+        this.inputStream.streamPos = streamPos;
```
rpastrana, line 185: are negative numbers ignored?

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
```diff
@@ -138,10 +149,41 @@ public HpccRemoteFileReader(DataPartition dp, FieldDef originalRD, IRecordBuilde
...
+        {
+            RowServiceInputStream.RestartInformation restartInfo = new RowServiceInputStream.RestartInformation();
+            restartInfo.streamPos = resumeInfo.inputStreamPos;
+            restartInfo.tokenBin = resumeInfo.tokenBin;
+
+            this.inputStream = new RowServiceInputStream(this.dataPartition, this.originalRecordDef, projectedRecordDefinition, connectTimeout, limit, createPrefetchThread, readSizeKB, restartInfo);
+            long bytesToSkip = resumeInfo.recordReaderStreamPos - resumeInfo.inputStreamPos;
+            if (bytesToSkip < 0)
+            {
+                throw new Exception("Unable to restart unexpected stream pos in record reader.");
```
rpastrana, line 168: would it be helpful at all to output the stream pos as part of the exception?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -290,6 +312,24 @@ public void run()
...
 
+    RestartInformation getRestartInformationForStreamPos(long streamPos)
+    {
+        RestartInformation restartInfo = new RestartInformation();
+
+        for (int i = 0; i < streamPosOfFetches.size(); i++)
+        {
+            int index = streamPosOfFetches.size() - (i+1);
+            Long fetchStreamPos = streamPosOfFetches.get(index);
+            if (fetchStreamPos <= streamPos) {
```
rpastrana, line 323: very minor format issue

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -290,6 +312,24 @@ public void run()
         }
     }
 
+    RestartInformation getRestartInformationForStreamPos(long streamPos)
+    {
+        RestartInformation restartInfo = new RestartInformation();
+
```
rpastrana, line 318: does this simplify this code a little bit? for (int index = streamPosOfFetches.size() - 1 ; index > -1; index--)

rpastrana (commented):
@jpmcmu looks good, left a couple of minor comments. As you mentioned, junit should be added.

jpmcmu:
@rpastrana code review changes have been made and junit has been added. Please review again

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
jpmcmu: Good question, I don't think it would help track down why the resume info wasn't valid. This is essentially a sanity check

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -290,6 +335,31 @@ public void run()
...
+     * Returns RestartInformation for the given streamPos if possible.
+     * 
+     * @return RestartInformation
+     */
+    RestartInformation getRestartInformationForStreamPos(long streamPos)
+    {
+        RestartInformation restartInfo = new RestartInformation();
+
+        // Note if we don't find a valid start point we will restart from the beginning of the file
+        for (int i = 0; i < streamPosOfFetches.size(); i++)
```
rpastrana, line 348: It seems we can simplify this for loop condition having a lone index, and having it decrement.
At the very least, let's move the streamPosOfFetches.size() calls out of the loop condition and body onto a local var.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -67,6 +67,8 @@ public void readBadlyDistributedFileTest() throws Exception
         //this file only has data on two nodes
         HPCCFile file = new HPCCFile(datasets[1], connString , hpccUser, hpccPass);
         file.setProjectList("");
+        file.setFileAccessExpirySecs(10);
```
rpastrana, line 70: These values are always tough to predict, why not make it configurable?
Also, is 10 secs meant to ensure we don't time out? If so, how close do we usually come to expiring before finishing?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -402,6 +413,129 @@ public void filteredTest() throws Exception
...
+            // Create a new file reader with the resume info from above
+            HpccRemoteFileReader<HPCCRecord> fileReader = new HpccRemoteFileReader<HPCCRecord>(fileParts[i], originalRD, recordBuilder,
+                                                                                                -1, -1, true, -1, partitionResumeList.get(i));
+
+            while (fileReader.hasNext())
+            {
+                HPCCRecord record = fileReader.next();
+                if (record == null)
+                {
+                    Assert.fail("Received null record during read");
```
rpastrana, line 503: Received null record during RESUME read

rpastrana (commented):
@jpmcmu looks good, but a couple of minor comments

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: I changed the timeout to reduce the time it takes for the tests to run. At the moment in readFile() there is code that purposely causes a timeout to test if reading can continue on active connections after the timeout is up. That is one of the reasons the full suite of tests was taking a while to run. I reduced the timeout from 120s -> 10s to speed up the tests.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: Good call

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: Removing those changes for now. Will address in a separate PR

rpastrana (approved):
Approved

PR #486, opened 2021-04-27
JAPI-408 Address array out of bounds in DFSclient stream restart

jpmcmu:
- Fixed issue with out of bounds issue in getResumeInfo
- Added unit test for resume functionality

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
@@ -428,15 +428,18 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     {
         RestartInformation restartInfo = new RestartInformation();
 
-        // Note if we don't find a valid start point we will restart from the beginning of the file
-        for (int i = streamPosOfFetches.size()-1; i >= 0; i--)
+        synchronized (streamPosOfFetches)
         {
-            Long fetchStreamPos = streamPosOfFetches.get(i);
-            if (fetchStreamPos <= streamPos)
+            // Note if we don't find a valid start point we will restart from the beginning of the file
+            for (int i = streamPosOfFetches.size()-1; i >= 0; i--)
             {
-                restartInfo.streamPos = fetchStreamPos;
-                restartInfo.tokenBin = tokenBinOfFetches.get(i);
-                break;
+                Long fetchStreamPos = streamPosOfFetches.get(i);
+                if (fetchStreamPos <= streamPos)
+                {
+                    restartInfo.streamPos = fetchStreamPos;
+                    restartInfo.tokenBin = tokenBinOfFetches.get(i);
+                    break;
+                }
             }
         }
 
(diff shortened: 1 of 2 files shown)


jpmcmu:
@rpastrana @drealeed please review

drealeed (approved):
Looks good to me

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -956,19 +962,18 @@ private void finishFetch()
...
-
             if (this.tokenBin == null || tokenLen > this.tokenBin.length)
             {
                 this.tokenBin = new byte[tokenLen];
             }
-            else
+            dis.readFully(this.tokenBin,0,tokenLen);
+            
+            this.streamPosOfFetchStart += totalDataInCurrentRequest;
+            synchronized (streamPosOfFetches)
```
rpastrana, line 972: is it possible for "this.tokenBin" to be empty or invalid at this point? Is there any way to confirm the token?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -126,6 +126,77 @@ public void integrationReadWriteBackTest() throws Exception
...
+            Assert.fail("No file parts found");
+        }
+
+        FieldDef originalRD = file.getRecordDefinition();
+        if (originalRD == null || originalRD.getNumDefs() == 0)
+        {
+            Assert.fail("Invalid or null record definition");
+        }
+
+        ArrayList<HpccRemoteFileReader.FileReadResumeInfo> resumeInfo = new ArrayList<HpccRemoteFileReader.FileReadResumeInfo>();
```
rpastrana, line 146: outside the scope of this issue, but should we consider providing a resume info container class with the filepart id and the array of FileReadResumeInfo?

rpastrana (commented):
@jpmcmu looks good, just a couple of questions.

rpastrana:
@drealeed can you verify this fix works for you before we merge it? thanks.

drealeed:
I'll check it out tomorrow and let you know

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
jpmcmu: If we are given an invalid token at this point it would be a significant issue. I don't think there is a good recovery method in that case. We do have logging if a token is rejected so we would be able to trace the issue.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: Good question, in an actual use case this wouldn't be required as there would be a one-to-one mapping between the FileReadResumeInfo and the file part. We need it here due to the test restarting the read after each record.

rpastrana (approved):
@jpmcmu looks good. We'll merge once we get confirmation from Drea

PR #559, opened 2022-05-18
JAPI-458 : New search files method using wrapper

drealeed:
## Type of change:

## Checklist:
  - [x] I have performed unit tests to cover my changes.
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
@@ -61,10 +61,7 @@ import org.hpccsystems.ws.client.utils.Connection;
 import org.hpccsystems.ws.client.wrappers.ArrayOfEspExceptionWrapper;
 import org.hpccsystems.ws.client.wrappers.EclRecordWrapper;
 import org.hpccsystems.ws.client.wrappers.EspSoapFaultWrapper;
-import org.hpccsystems.ws.client.wrappers.gen.wsdfu.AddtoSuperfileRequestWrapper;
-import org.hpccsystems.ws.client.wrappers.gen.wsdfu.AddtoSuperfileResponseWrapper;
-import org.hpccsystems.ws.client.wrappers.gen.wsdfu.DFUSearchDataRequestWrapper;
-import org.hpccsystems.ws.client.wrappers.gen.wsdfu.DFUSearchDataResponseWrapper;
+import org.hpccsystems.ws.client.wrappers.gen.wsdfu.*;
 import org.hpccsystems.ws.client.wrappers.wsdfu.DFUCreateFileWrapper;
 import org.hpccsystems.ws.client.wrappers.wsdfu.DFUDataColumnWrapper;
 import org.hpccsystems.ws.client.wrappers.wsdfu.DFUFileAccessInfoWrapper;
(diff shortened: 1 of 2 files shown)


wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -1297,6 +1294,52 @@ public List<DFULogicalFileWrapper> searchFiles(String logicalFilename, String cl
...
+     * @param request - the search request
+     * @return - collection of files matching the request passed in
+     * @throws java.lang.Exception general exception
+     * @throws org.hpccsystems.ws.client.wrappers.ArrayOfEspExceptionWrapper array of esp exception wrapper
+     */
+    public List<DFULogicalFileWrapper> searchFiles(DFUQueryRequestWrapper request) throws Exception, ArrayOfEspExceptionWrapper
+    {
+        verifyStub(); // Throws exception if stub failed
+
+        if (request.getLogicalName() != null && request.getLogicalName().startsWith("~"))
```
rpastrana, line 1310: If there's ever a situation where the leading tilde is necessary, we should add an input param to this method to allow caller to include the tilde
Also, since we are changing the caller's input, we should report this to the logger

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
```diff
@@ -1297,6 +1294,52 @@ public List<DFULogicalFileWrapper> searchFiles(String logicalFilename, String cl
...
+            handleEspSoapFaults(new EspSoapFaultWrapper(e), "Could Not SearchFiles");
+        }
+
+        if (resp.getExceptions() != null) handleEspExceptions(new ArrayOfEspExceptionWrapper(resp.getExceptions()), "Could Not SearchFiles");
+
+        List<DFULogicalFileWrapper> result = new ArrayList<DFULogicalFileWrapper>();
+        ArrayOfDFULogicalFile logicalfilearray = resp.getDFULogicalFiles();
+        if (logicalfilearray != null)
+        {
+            DFULogicalFile[] dfulogicalfilearray = resp.getDFULogicalFiles().getDFULogicalFile();
```
rpastrana, line 1334: we could use logicalfilearray  instead of calling resp.getDFULogicalFiles() again

rpastrana (changes requested):
@drealeed looks good, just a couple of minor comments.

wsclient/src/test/java/org/hpccsystems/ws/client/WsDFUClientTest.java
```diff
@@ -360,6 +361,16 @@ public void testDeleteFileWithprefixAtsignCluster() throws Exception
     public void searchFileTest() throws Exception
     {
         String searchStr="hipie::keys::*";
+
+
+        DFUQueryRequestWrapper request=new DFUQueryRequestWrapper();
+        request.setFileType("Superfiles Only");
```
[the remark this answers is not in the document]
drealeed: Tell me about it...I had to open up the debug network console in ecl watch and examine the actual value being sent when I filtered by superfiles to figure out what value to use.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
drealeed: this is being done in all the search methods--if we change it here we should also change it in the other search methods. Separate PR?

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
drealeed: Fixed

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsDFUClient.java
rpastrana: Gotcha. It feels like the previous method should populate a DFUQueryRequestWrapper and call into yours (to avoid replicated logic). We can handle that as an improvement in a separate PR. https://track.hpccsystems.com/browse/JAPI-460

rpastrana (approved):

PR #583, opened 2022-08-17
HPCC4J-477 Ensure DZ paths contain trailing slash > 7.12.98

rpastrana:
- Provide utilities to ensure trailing slash on a given path string
- Provide junit for ensure trailing slash utility
- Ensure all DZ paths acquired from hpcc >= 7.12.98 contain trailing slash
- Provide DZ path junit

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCFileSprayClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCFileSprayClient.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Properties;
 
 import javax.xml.parsers.DocumentBuilder;
-import javax.xml.parsers.DocumentBuilderFactory;
 import javax.xml.parsers.ParserConfigurationException;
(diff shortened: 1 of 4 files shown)


rpastrana:
@jpmcmu please review

jpmcmu (approved):
@rpastrana looks good

jpmcmu:
@rpastrana Looks good, it does seem like we need to figure out a global method to handle the separators, but if I am understanding correctly this isn't something we can do it atm.

rpastrana:
> @rpastrana Looks good, it does seem like we need to figure out a global method to handle the separators, but if I am understanding correctly this isn't something we can do it atm.

Well the remote DZ path separator has to communicated to the client. As far as I can tell we have to infer the separator from the "isLinux" response field. I'm not sure if that addresses your concern... but remember the path is not local so the client cannot assume it uses the local file system separator.

PR #588, opened 2022-08-23
HPCC4J-487 Add support for WsCloud

rpastrana:
- Adds WsCloud WSDL
- Adds Axis2 generated stub and corresponding mvn profile
- Adds corresponding junit

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsCloudClient.java
+++ b/wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsCloudClient.java
@@ -1,0 +1,270 @@
+/*##############################################################################
+
+    HPCC SYSTEMS software Copyright (C) 2022 HPCC Systems®.
+
+    Licensed under the Apache License, Version 2.0 (the "License");
+    you may not use this file except in compliance with the License.
+    You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+    Unless required by applicable law or agreed to in writing, software
+    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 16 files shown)


rpastrana:
@jpmcmu please review

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsCloudClient.java
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    /**
+     * <p>getPods.</p>
+     *
+     * @return a JSON formatted list of available pods - in raw k8s format.
+     * @throws java.lang.Exception if any.
+     */
+    public String getPods() throws Exception
```
jpmcmu, line 233: Returning raw k8s format here is interesting; I don't think we should take on providing consistency here, but it does lead me to wonder what the use cases for this are?

jpmcmu (approved):
@rpastrana Looks good, question about the raw k8s format, but no real issues.

wsclient/src/main/java/org/hpccsystems/ws/client/HPCCWsCloudClient.java
rpastrana: Agreed, I'm a bit surprised the ESP service was designed this way as well, typically they provide well defined response structures.
The use case is likely related to the new ECLWatch interface which lists hpcc cluster pods, and also uses those pod names as potential filters in the new log viewer.
As far as wsclient, this is added for completeness.

PR #606, opened 2023-03-31
JAPI-507 DFSClient: Add support for new rowservice protocol

jpmcmu:
## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
@@ -51,6 +51,7 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     private boolean                  simulateFail = false;
     private boolean                  forceTokenUse = false;
     private boolean                  inFetchingMode = false;
+    private boolean                  useOldProtocol = true;
     private byte[]                   tokenBin;
     private int                      handle;
     private DataPartition            dataPart;
@@ -58,8 +59,10 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     private String                   jsonRecordDefinition          = null;
     private FieldDef                 projectedRecordDefinition     = null;
     private String                   projectedJsonRecordDefinition = null;
-    private java.io.DataInputStream  dis;
-    private java.io.DataOutputStream dos;
+    private java.io.DataInputStream  dis = null;
+    private java.io.DataOutputStream dos = null;
+
+    private String                   rowServiceVersion = "";
 
     private int                      filePartCopyIndexPointer = 0;  //pointer into the prioritizedCopyIndexes struct
     private List<Integer>            prioritizedCopyIndexes = new ArrayList<Integer>();
@@ -81,8 +84,8 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     private int                      totalDataInCurrentRequest = 0;
     private int                      remainingDataInCurrentRequest = 0;
     private long                     streamPos = 0;
-    
-    // Used for restarts 
+
+    // Used for restarts
     private long                     streamPosOfFetchStart = 0;
     private List<Long>               streamPosOfFetches = new ArrayList<Long>();
     private List<byte[]>             tokenBinOfFetches = new ArrayList<byte[]>();
@@ -108,6 +111,7 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     private int                      connectTimeout = DEFAULT_CONNECT_TIMEOUT_MILIS;
 
     private static final Charset     HPCCCharSet                   = Charset.forName("ISO-8859-1");
+    private static final Logger      log                           = LogManager.getLogger(RowServiceInputStream.class);
 
     // Note: The platform may respond with more data than this if records are larger than this limit.
     private static final int         DEFAULT_MAX_READ_SIZE_KB = 4096;
@@ -118,8 +122,6 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     // the network connection is slow. The read on the socket will block until
     // at least 512 bytes are available
     private static final int         MIN_SOCKET_READ_SIZE     = 512;
-
-    private static final Logger      log                           = LogManager.getLogger(RowServiceInputStream.class);
 
     private int maxReadSizeKB = DEFAULT_MAX_READ_SIZE_KB;
 
@@ -147,6 +149,14 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     {
         public long streamPos = 0;
         public byte[] tokenBin = null;
+    }
+
+    private static class RowServiceResponse
+    {
+        int len = 0;
+        int errorCode = 0;
+        int handle = -1;
+        String errorMessage = null;
     }
 
     /**
@@ -275,9 +285,9 @@ public class RowServiceInputStream extends InputStream implements IProfilable
      *            Wether or not this inputstream should handle prefetching itself or if prefetch will be called externally
      * @param maxReadSizeInKB
      *            max readsize in kilobytes
-     * @param restartInfo 
+     * @param restartInfo
      *            information used to restart a read from a particular stream position
-     * @param isFetching 
+     * @param isFetching
      *            Will this input stream be used to serviced batched fetch requests
      * @throws Exception
      *            general exception
@@ -423,7 +433,7 @@ public class RowServiceInputStream extends InputStream implements IProfilable
 
     /**
      * Returns RestartInformation for the given streamPos if possible.
-     * 
+     *
      * @return RestartInformation
      */
     RestartInformation getRestartInformationForStreamPos(long streamPos)
@@ -448,11 +458,6 @@ public class RowServiceInputStream extends InputStream implements IProfilable
         return restartInfo;
     }
 
-    /**
-     * Sets the next file part copy.
-     *
-     * @return true, if successful
-     */
     private boolean setNextFilePartCopy()
     {
         if (filePartCopyIndexPointer + 1 >= prioritizedCopyIndexes.size()) return false;
@@ -481,11 +486,6 @@ public class RowServiceInputStream extends InputStream implements IProfilable
         return this.dataPart.getCopyIP(prioritizedCopyIndexes.get(getFilePartCopy()));
     }
 
-    /**
-     * The current file part's target copy index.
-     *
-     * @return Current filepart copy index (0-indexed)
-     */
     private int getFilePartCopy()
     {
         return filePartCopyIndexPointer;
@@ -499,16 +499,6 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     public int getPort()
     {
         return this.dataPart.getPort();
-    }
-
-    /**
-     * The read transaction in JSON format.
-     *
-     * @return read transaction
-     */
-    public String getTrans()
-    {
-        return this.makeInitialRequest();
     }
 
     /**
@@ -529,6 +519,16 @@ public class RowServiceInputStream extends InputStream implements IProfilable
     public boolean isClosed()
     {
         return this.closed.get();
+    }
+
+    /**
+     * Is the remote rowservice using the new message protocol introduced in 8.12.10?
+     *
+     * @return true if using new protocol
+     */
+    public boolean isUsingNewProtocol()
+    {
+        return !this.useOldProtocol;
     }
 
     /**
@@ -592,7 +592,7 @@ public class RowServiceInputStream extends InputStream implements IProfilable
 
     /**
      * Starts a block fetching request for the requested records.
-     * 
+     *
      * Starts a block fetching request that will block until all of the requested records have been consumed.
      * Should only be called from a separate thread from the thread that is consuming data from the input stream.
      * @param fetchOffsets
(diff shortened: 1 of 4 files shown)


jpmcmu:
@rpastrana please review

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -182,26 +203,64 @@ public class RowServiceOutputStream extends OutputStream
             throw new Exception(errorMessage);
         }
 
+        //------------------------------------------------------------------------------
+        // Check protocol version
+        //------------------------------------------------------------------------------
+
+        try
+        {
+            String msg = makeGetVersionRequest();
```
rpastrana, line 212: what's the expected return format, and do we log the actual interface version used?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -222,79 +281,167 @@ private void makeInitialWriteRequest() throws Exception
...
+
+        try
         {
-            throw new IOException("Received short or truncated response from row service. Aborting.");
+            this.dos.writeInt(jsonRequestLen + 4 + 1);
+            this.dos.write((int)RFCCodes.RFCStreamGeneral);
+            this.dos.writeInt(jsonRequestLen);
+            this.dos.write(closeFileRequest.getBytes(StandardCharsets.ISO_8859_1));
+            this.dos.flush();
+            System.out.println("Sent close request for handle: " + this.handle);
```
rpastrana, line 321: Do we have a logger available?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -86,7 +88,7 @@ public void readWithForcedTimeoutTest() throws Exception
     @Test
     public void integrationReadWriteBackTest() throws Exception
     {
-        for (int i = 0; i < datasets.length; i++)
+        for (int i = 0; i < 1; i++)
```
rpastrana, line 91: are we skipping the second dataset test?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -946,93 +1012,107 @@ public List<HPCCRecord> readFile(HPCCFile file, Integer connectTimeoutMillis, bo
 
         return records;
     }
-
+
```
rpastrana, line 1015: extra whitespace?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
```diff
@@ -946,93 +1012,107 @@ public List<HPCCRecord> readFile(HPCCFile file, Integer connectTimeoutMillis, bo
...
+    }
+    
+    private void writeFile(List<HPCCRecord> records, String fileName, FieldDef recordDef, Integer connectTimeoutMs, boolean compressed)
     {
         try
         {
-            //------------------------------------------------------------------------------
-            //  Request a temp file be created in HPCC to write to
-            //------------------------------------------------------------------------------
+            writeFileWithExceptions(records, fileName, recordDef, connectTimeoutMs, compressed);
```
rpastrana, line 1025: is this test expecting to cause an exception, if so, the catch should fail, right?

rpastrana (changes requested):
@jpmcmu a few comments/questions.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -738,114 +737,87 @@ private int startFetch()
...
-        }
-
-        // If len == 0 we have finished reading the file
-        if (len == 0)
-        {
-            try
+            catch (HpccFileException e)
             {
-                close();
+                prefetchException = e;
```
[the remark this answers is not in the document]
jpmcmu: It will get handled, but since that code can be executed by the prefetch thread we postpone handling the exception until the reading thread can process it.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -738,114 +737,87 @@ private int startFetch()
...
+                {
+                    close();
+                }
+                catch(Exception ie){}
+                return -1;
             }
-            catch (IOException e)
+
+            // If len < 0 we have finished reading the file
+            if (response.len < 0)
```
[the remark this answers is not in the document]
jpmcmu: Response cannot be null here.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -738,114 +737,87 @@ private int startFetch()
...
 
-        // Invalid response len, should be at least 4
-        if (len < 4)
-        {
-            prefetchException = new HpccFileException("Early data termination, no handle");
-            try
+            if (response.errorCode != RFCCodes.RFCStreamNoError)
             {
-                close();
+                prefetchException = new HpccFileException(response.errorMessage);
```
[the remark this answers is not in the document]
jpmcmu: Good question, there response error message includes the error code.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -738,114 +737,87 @@ private int startFetch()
...
+            //------------------------------------------------------------------------------
+            // Retry with the token if handle is invalid
+            //------------------------------------------------------------------------------
 
-        try
-        {
-            this.handle = dis.readInt();
-            if (this.handle == 0)
+            this.handle = response.handle;
+            if (this.handle == 0 && !inTokenRetry)
```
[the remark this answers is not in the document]
jpmcmu: The handle at this point will always either be 0 or greater than 0, but it would probably be better to change that to this.handle <= 0; In case that changes in the future.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -738,114 +737,87 @@ private int startFetch()
...
-                    catch(Exception ie){}
+                    prefetchException = new HpccFileException("Failed on remote read read retry", e);
                     return -1;
                 }
-
-                if (len == 0)
+            }
+            else if (this.handle == 0)
+            {
+                prefetchException = new HpccFileException("Read retry failed");
```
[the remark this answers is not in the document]
jpmcmu: The if statement above will be executed if we are in a retry, so that exception will only be thrown during a retry.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -1491,118 +1465,142 @@ public List<IMetric> getMetrics()
...
+                this.dos.writeInt(msgLen);
+                this.dos.write(msg.getBytes(HPCCCharSet), 0, msgLen);
+                this.dos.flush();
+            }
+            catch (IOException e)
+            {
+                throw new HpccFileException("Failed on initial remote read read trans", e);
+            }
+
+            RowServiceResponse response = response = readResponse();
```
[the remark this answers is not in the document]
jpmcmu: Oh yeah, definitely a typo good catch.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -182,26 +203,64 @@ public class RowServiceOutputStream extends OutputStream
...
+        //------------------------------------------------------------------------------
+        // Check protocol version
+        //------------------------------------------------------------------------------
+
+        try
+        {
+            String msg = makeGetVersionRequest();
+            int msgLen = msg.length();
+
+            this.dos.writeInt(msgLen);
```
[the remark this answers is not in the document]
jpmcmu: There would be a slight increase in connect time, but it should be pretty close to the ping time. So on internal networks it would only add < 10ms? Which I think it negligible

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
```diff
@@ -1860,18 +1879,25 @@ private void makeFetchObject(StringBuilder sb)
...
+    private String makeGetVersionRequest()
+    {
+        final String versionMsg = RFCCodes.RFCStreamReadCmd + "{ \"command\" : \"version\", \"handle\": \"-1\", \"format\": \"binary\" }";
+        return versionMsg;
+    }
+
     private String makeInitialRequest()
     {
-        StringBuilder sb = new StringBuilder(2048);
+        StringBuilder sb = new StringBuilder(256);
```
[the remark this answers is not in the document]
jpmcmu: It is an arbitrary value, that is the initial buffer size within the string buffer.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: At the moment I am not checking the version number, just if we got a valid response.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: We do, but that is left over from debugging. Will remove that, good catch

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: Left over from debugging will fix.

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: The emptyCompressedFileTest test is expecting an exception which is why I split that out into two functions. In that test I return if the exception is thrown so that the test succeeds and Assert.fail otherwise

jpmcmu:
@rpastrana Code review changes pushed up please review, will squash if it looks good.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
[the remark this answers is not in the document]
rpastrana: the else-if clause doesn't seem to be considered a retry (based on variables used, and your reply, but the error message in this clause says it is a failed retry). If the error message is correct, ignore, otherwise ensure the message is not misleading.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
rpastrana: do we know the expected response format? If so, let's document here (as a comment perhaps).
Can we extract the reported interface version? If so, let's capture it and make it available, or do we expect this version to be the last change?

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
rpastrana: Ok, I think the writeFileWithExceptions name is causing confusion. At face-value that function would seem to write a file which will cause exceptions, but it seems to write a file, and report Exceptions if encountered. A slight name adjustment to something like writeFileAndReportAnyExceptions would avoid confusion.

rpastrana (changes requested):
@jpmcmu see latest comments.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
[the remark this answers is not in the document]
jpmcmu: No, it is the same end-point. Yes, the legacy server does respond to that version request, but essentially with an "I don't understand this request response." Which is what I am checking for to determine which protocol to use. Jake explicitly designed that version request so that the legacy server would be able to respond with the I don't understand response.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: Yep, I do. Good point, I will add that

dfsclient/src/test/java/org/hpccsystems/dfs/client/DFSReadWriteTest.java
jpmcmu: Yeah, that is a better name. I will make that change.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceInputStream.java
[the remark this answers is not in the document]
jpmcmu: My original comment had a typo. I meant to say the "the if statement above will be executed if we aren't in a retry". So, the error message is correct. If we head a read failure we would take the first branch and move into the token retry, if that failed we would end up in the second branch.

` if (this.handle <= 0 && !inTokenRetry)`

jpmcmu:
@rpastrana code review changes pushed up. If it looks good will squash

rpastrana (approved):
approved

PR #615, opened 2023-06-07
HPCC4J-509 Update ESP 9.0.x Interfaces

rpastrana:
- Does not update wswu nor wsfs

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


wsclient/src/main/java/org/hpccsystems/ws/client/wrappers/gen/filespray/AbortDFUWorkunitWrapper.java
```diff
@@ -20,11 +20,11 @@
 
 /**
  * Generated Axis2 ADB stub class wrapper
- * WrapperMaker version: 1.7
+ * WrapperMaker version: 1.8
  * Class name: AbortDFUWorkunitWrapper
  * Wraps class: org.hpccsystems.ws.client.gen.axis2.filespray.latest.AbortDFUWorkunit
  * Output package : org.hpccsystems.ws.client.wrappers.gen.filespray
- * Service version: 1.23
```
jpmcmu, line 27: What are your thoughts on potentially removing the service version here? I think it would make these interface change commits a lot cleaner in the future, in that only the wrappers that have changes would be present in the commit. I think this could be beneficial in understanding what has changed in between service versions.

jpmcmu:
@rpastrana I am seeing some files in the PR from a commit last week, that aren't showing as changed when I run the interface generation tool locally. IE: ListUserIDsResponse.java is showing as changed in your commit on the 7th but isn't being touched when I run the interface generation tool locally. Maybe the version of the tool on the 7th was incorrectly updating those files?

jpmcmu (changes requested):
I am seeing a discrepancy between the tool output locally and the file changes here. It seems like the PR may include changes from older tool versions and might need to have the older commits removed from its history.

wsclient/src/main/java/org/hpccsystems/ws/client/wrappers/gen/filespray/AbortDFUWorkunitWrapper.java
rpastrana: I definitely would like to prevent unnecessary diffs, but the corresponding service version seems useful, even if the class didn't change.

rpastrana:
> @rpastrana I am seeing some files in the PR from a commit last week, that aren't showing as changed when I run the interface generation tool locally. IE: ListUserIDsResponse.java is showing as changed in your commit on the 7th but isn't being touched when I run the interface generation tool locally. Maybe the version of the tool on the 7th was incorrectly updating those files?

Per offline conversation, it seems the discrepancy had to do w/ the Axis tool timestamp and not an issue. The other issue we found are the changes in eol, we'll let this set of unintended eol changes through and seperately we'll standardize all eols and add a git rule to not automatically change eol: https://track.hpccsystems.com/browse/HPCC-29740
please re-review

jpmcmu (approved):

jpmcmu:
@rpastrana looks good to me

rpastrana:
@jpmcmu take a look at the latest commit, once we're good w/ the wrapper tool changes, we'll update all the other clients

jpmcmu (approved):
@rpastrana looks good to me

PR #667, opened 2023-12-14
HPCC4J-561 copyfile test wait for spray completion

jpmcmu:
- Modified WSFileIOClientTest.copyfile test to correctly monitor spray progress
- Updated remote unit tests values.yaml to keep file spray service

## Type of change:

## Checklist:
- [X] My change requires a change to the documentation.
  - [X] I have updated the documentation accordingly, or...
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
+++ b/wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
@@ -21,7 +21,9 @@ import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.fail;
 import static org.junit.Assume.assumeTrue;
 
+import java.util.Arrays;
 import java.util.HashSet;
+import java.util.List;
 import java.util.Set;
 
 import org.apache.axis2.AxisFault;
@@ -67,9 +69,6 @@ public class WSFileIOClientTest extends BaseRemoteTest
     @Test
     public void copyFile() throws Exception
     {
-        Assume.assumeFalse("Test not valid on containerized HPCC environment", client.isTargetHPCCContainerized());
-        assumeTrue("Ignoring test 'copyFile' because HPCC-30117 is not fixed", HPCC_30117.equalsIgnoreCase("fixed"));
-
         String lzfile=System.currentTimeMillis() + "_csvtest.csv";
         String hpccfilename="temp::" + lzfile;
         client.createHPCCFile(lzfile, targetLZ, true);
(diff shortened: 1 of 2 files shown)


jpmcmu:
@rpastrana please review

wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
```diff
@@ -67,45 +69,47 @@ public class WSFileIOClientTest extends BaseRemoteTest
...
-                        Thread.sleep(1000);
-                    }
+                    Thread.sleep(1000);
+                    System.out.println("Percent complete: " + dfuProgress.getPercentDone() + " Sleeping for 1sec to wait for spray.");
+                    waitCount++;
                 }
-            }
+            } while (waitCount < 60 && dfuProgress.getPercentDone() < 100);
+
+            assumeTrue("File spray never completed", waitCount < MAX_WAIT_COUNT);
```
rpastrana, line 112: Minor, perhaps picky, but the message might be a tiny bit misleading, we don't know if the file spray completed, but we do know the test timed out

wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
```diff
@@ -67,45 +69,47 @@ public class WSFileIOClientTest extends BaseRemoteTest
...
-                    if (dfuspray.getSecsLeft()==0)
-                    {
-                        i=0;
-                    }
-                    else
-                    {
-                        Thread.sleep(1000);
-                    }
+                    Thread.sleep(1000);
+                    System.out.println("Percent complete: " + dfuProgress.getPercentDone() + " Sleeping for 1sec to wait for spray.");
```
rpastrana, line 107: Minor because we're not likely to see this message often, but when we do, it feels incomplete because we didn't announce we're spraying a file and I'm not sure it'll be obvious why we're reporting status for a spray.

wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
```diff
@@ -67,45 +69,47 @@ public class WSFileIOClientTest extends BaseRemoteTest
...
+            int waitCount = 0;
+            int MAX_WAIT_COUNT = 60;
+
+            ProgressResponseWrapper dfuProgress = null;
+            do
             {
-                System.out.println("Still spraying, waiting 1 sec...");
-                for (int i=wait;i>0;i--)
+                dfuProgress = wsclient.getFileSprayClient().getDfuProgress(dfuspray.getWuid());
+                if (!whiteListedStates.contains(dfuProgress.getState())) {
```
rpastrana, line 100: Since we're testing for a failure here, it feels more natural to me to seek for one of any blacklisted states instead. Is there something odd about the failure states that swayed you to seek for absence of whitelisted states?

rpastrana (changes requested):
@jpmcmu just a few questions.

wsclient/src/test/java/org/hpccsystems/ws/client/WSFileIOClientTest.java
jpmcmu: I generally go for whitelists because they are explicit and more likely to catch errors / future errors. IE: If a new state gets added the whitelist will treat it as an error but the blacklist would ignore it.

rpastrana: Fair enough, but the original enum on the platform side doesn't seem to have changed in a while:
https://github.com/hpcc-systems/HPCC-Platform/blame/7552030a5bb03b63d4af0d37d454f8eab22feb9b/dali/dfu/dfuwu.hpp#L40
If that is the right enum, at this point I would consider it static.

With that in mind, it seems seeking explicitly for one of the known failed states will lead to less false positives.

jpmcmu:
@rpastrana Pushed up code review changes & squashed

rpastrana (approved):
@jpmcmu approved.

PR #679, opened 2024-01-31
JAPI-573 - Add in default support for set of 

jchambers-ln:
https://track.hpccsystems.com/browse/JAPI-573

## Type of change:

## Checklist:
  - [x] This change does not cause any existing JUnits to fail.
  - [x] I have include JUnit coverage to test this change
  - [x] I have performed system test and covered possible regressions and side effects.
  - [x] Scalability
  - [x] Performance
  - [x] Security
  - [x] Thread-safety
  - [x] Premature optimization
  - [x] This change fixes the problem, not just the symptom

## Testing:
Basic unit test added to hpcc4j, additional testing added to DSP to validate it works with the requirements leading to it being added.


wsclient/src/test/java/org/hpccsystems/ws/client/DFUFileDetailWrapperTest.java
```diff
@@ -332,6 +332,34 @@ public void testGetRecordMaxlength() throws Exception {
         assertNotNull(column);
     }
 
+    @Test
+    public void testSetDefault() throws Exception {
+        String setof = "RECORD\nSET OF UTF8 field3set {DEFAULT ([''])};\nEND;";
+        EclRecordWrapper info = DFUFileDetailWrapper.getRecordEcl(setof);
```
drealeed, line 338: add this to the constants at the top of this file. Also add a field before and after the SET in the record and ensure that they are still handled as expected

jchambers-ln: updated

jpmcmu (approved):
@jchambers-ln looks good to me

jpmcmu:
@jchambers-ln Can you squash and retarget 9.4.x?

wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecord.g4
```diff
@@ -104,6 +105,10 @@ xmldefaultval:
     'XMLDEFAULT' OPAREN STRING CPAREN
 ;
 
+setdefaultval:
+    ('DEFAULT' OPAREN OSQUARE STRING CSQUARE CPAREN | 'DEFAULT' OPAREN SETSTRING CPAREN)
```
rpastrana, line 109: @jchambers-ln what's the difference between 
'DEFAULT' OPAREN OSQUARE STRING CSQUARE CPAREN
and 
'DEFAULT' OPAREN SETSTRING CPAREN

jpmcmu:
@jchambers-ln Actually since this issue you are addressing could happen in older versions it make sense to back port your fix. If possible retarget to 8.12.x instead of 9.4.x, if you find that there are conflicts that make back porting it that far problematic let me know.

rpastrana:
@ghalliday I just noticed our ECL record definition parser grammar doesn't seem to be in synch with the latest doc:

image

How can we get guidance on what syntax should be supported?

rpastrana:
@jchambers-ln please squash and re-target, and address this question: https://github.com/hpcc-systems/hpcc4j/pull/679#discussion_r1478400208

rpastrana:
@ghalliday please comment on the overall grammar correctness.

wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecord.g4
ghalliday, line 109: I think

`
(
'DEFAULT' OPAREN OSQUARE value_list CSQUARE CPAREN |
'DEFAULT' OPAREN OSQUARE CSQUARE CPAREN
)
`
would be a more accurate rule.  It would allow for lists of integers.  The second production is for the empty list (which you will want to implement differently).  Note you could also have the keyword ALL, but I suspect that is very unlikely and I think you are only aiming to cover the common cases.

wsclient/src/test/java/org/hpccsystems/ws/client/DFUFileDetailWrapperTest.java
```diff
@@ -34,6 +34,12 @@
 @Category(org.hpccsystems.commons.annotations.BaseTests.class)
 public class DFUFileDetailWrapperTest
 {
+    private static final String SETOF = "RECORD\n" +
+            "STRING1 test;\n" +
+            "SET OF UTF8 field3set {DEFAULT ([''])};\n" +
+            "SET OF UTF8 field4set {DEFAULT (['test'])};\n" +
```
ghalliday, line 40: Test case would benefit from i) empty set ii) more than one string in the list  iii) integer values in the list (for a set of integer).

wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecordReader.java
```diff
@@ -396,6 +396,30 @@ else if (currentrec != null)
...
+     *
+     * <p>
+     * When entering a DEFAULT option for Set Of, set the columnvalue of the current field/rec to its value
+     * </p>
+     */
+    @Override
+    public void enterSetdefaultval(EclRecordParser.SetdefaultvalContext ctx)
+    {
+        String val = ctx.getChild(2).getText();
+        val = val.replace("['", "");
```
ghalliday, line 410: What happens if the string contains a square bracket?
e.g.  DEFAULT(['[]').

ghalliday (commented):
A few comments.

Also, if there is an option to generate a case-insensitive parser then it would be worth enabling that - to simplify the maxlength rule.  I think it is possible for any case combinations to occur for MAXLENGTH,MAXCOUNT,XPath etc - because the case is preserved.

rpastrana:
@jchambers-ln any update on this?

jchambers-ln:
I haven't had time to circle back, I'm no expert in the antlr parser.  The change I made isn't eloquent, but the definition as it is built is pretty muddy so would to make it cleaner the entire grammar would need to be reworked.

wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecordReader.java
jchambers-ln: well ['[]'] will not match [' or '] so you would end up with a default value of [ -- we could shift it to replace first and replace last this would only be an issue with DEFAULT(['['']) which doesn't make sense as you would have to escape the ' and it would need to be DEFAULT(['[\'']) and [\' wouldn't match the replacement

wsclient/src/test/java/org/hpccsystems/ws/client/DFUFileDetailWrapperTest.java
jchambers-ln: This is only a partial solution to cover the cases we need, specifically the DEFAULT (['']) case. It just happened to support the single value also.  This solution will not support more than one value in the list.  I agree the grammar should support any possible value DEFAULT can but... the grammar would need to be completely reworked to handle that we don't have anyone working in antlr currently.

wsclient/src/main/java/org/hpccsystems/ws/client/antlr/EclRecord.g4
jchambers-ln: 'DEFAULT' OPAREN OSQUARE STRING CSQUARE CPAREN matches DEFAULT ([STRING])
'DEFAULT' OPAREN SETSTRING CPAREN matches DEFAULT(STRING)

I tried several variations this is the one that worked, if you have someone that knows antlr better I'm open to suggestions.

rpastrana:
@jchambers-ln if we want this in the 9.4.x stream, it needs to be re-targeted.

jchambers-ln:
@rpastrana I opened a pull for 9.4.x https://github.com/hpcc-systems/hpcc4j/pull/697/files 

I was also able to rework the antlr to support all of the following


RECORD
    STRING1 test;
    SET OF UTF8 fieldAllset {DEFAULT (ALL)};
    SET OF UTF8 field2set {DEFAULT (['test1','test2'])};
    SET OF UTF8 field3set {DEFAULT ([''])};
    SET OF UTF8 field4set {DEFAULT (['test'])};
    STRING2 test;
 END;

PR #710, opened 2024-05-17
HPCC4J-594 Add support for skipping tests

jpmcmu:
- Updated version of surefire plugin to support excludes file
- Added logic to generate an excludes file to GH actions

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


drealeed:
another issue is the ecl generation code that generates the files failing for some reason against the 160 and other clusters. Rodrigo had me comment it out locally in order to get my tests to run. Do you know if your changes will address that too?

rpastrana:
Hi @drealeed those aren't necessarily related to this change, can you write up the executeECLScript problem? Thanks

dfsclient/pom.xml
```diff
@@ -34,6 +34,7 @@
                 <value>org.hpccsystems.ws.client.TestResultNotifier</value>
               </property>
             </properties>
+            <excludesFile>../excluded-tests.txt</excludesFile>
```
rpastrana, line 37: it seems ok to have a single excluded-tests file at the hpcc4j level, but I could also see an argument for dedicated excluded-tests files per sub project.
what are the main drawbacks/advantages of the current implementation?

pom.xml
```diff
@@ -45,7 +45,7 @@
     <maven.deploy.version>2.8.2</maven.deploy.version>
```
rpastrana, line 45: can we have a sample entry in the above excluded-tests.txt file?
Perhaps create a new test simply for the purpose of always excluding it via this file?
Also, does the github action mechanism append excluded tests entries, or does it overwrite the file?

rpastrana (commented):
@jpmcmu a few questions.

wsclient/src/test/java/org/hpccsystems/ws/client/platform/RampsDevRegressionTest.java
```diff
@@ -6,9 +6,11 @@
 
 import org.hpccsystems.ws.client.wrappers.ArrayOfEspExceptionWrapper;
 import org.hpccsystems.ws.client.wrappers.wsdfu.DFUDataColumnWrapper;
+import org.junit.Ignore;
 import org.junit.Test;
 import org.junit.experimental.categories.Category;
 
+@Ignore
```
rpastrana, line 13: @drealeed I'd like to integrate these tests into the standard wsclient tests, is there anything RAMPS specific being tested? If so, could you provide ECL to recreate the structures being testsed here?

drealeed: No, this was just a way to get a really comprehensive test of all possible record structures; I picked ramps dev because it had a lot of files for the regression test and I had access to it.  And at the time the ecl errors occurred reading files from ramps dev so I pointed the test at the previously failing files to make sure they worked now.

drealeed: Unfortunately the files tested in testFileMetadata() no longer exist. That test can be removed. We have similar tests in tardis for different hpcc ecl structures now.

rpastrana: Unless those ecl structures are proprietary, I'd like to absorb as many tests as possible. Does it make sense to donate those tests given the fact we can preemptively create the target files via ecl scripts?

rpastrana: @drealeed do you recall how test::wsclient::noecl was created? it seems like an edge case we should continue to test

PR #711, opened 2024-05-22
HPCC4J-604 Suppress EclWatch redirects

rpastrana:
- Appends rawxml param to all axis2 client calls

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


rpastrana:
@drealeed please take a look. I'm not convinced this change is 100% responsible for fixing the issue we were experiencing but the issue is no longer reproducible.

drealeed (approved):
It looks like that should fix the error I was getting. Thanks for jumping on it so fast.

rpastrana:
@jpmcmu please take a look.

wsclient/src/main/java/org/hpccsystems/ws/client/BaseHPCCWsClient.java
```diff
@@ -574,6 +574,10 @@ protected Integer getStubConnectionTO() throws AxisFault
     static public Stub setStubOptions(Stub thestub, Connection connection) throws AxisFault
     {
         Options opt = thestub._getServiceClient().getOptions();
+        EndpointReference toRef = opt.getTo();
+        String toAddress = toRef.getAddress() + (toRef.getAddress().contains("?") ? "&" : "?") + "rawxml_";
```
jpmcmu, line 578: @rpastrana I looked around to see if there was a different way to set a query parameter in Axis2, but wasn't able to find another method. However, I do wonder if we should be using the URL & URI classes here to add the query param instead of doing it manually.

wsclient/src/main/java/org/hpccsystems/ws/client/BaseHPCCWsClient.java
```diff
@@ -574,6 +574,10 @@ protected Integer getStubConnectionTO() throws AxisFault
     static public Stub setStubOptions(Stub thestub, Connection connection) throws AxisFault
     {
         Options opt = thestub._getServiceClient().getOptions();
+        EndpointReference toRef = opt.getTo();
```
jpmcmu, line 577: Lets add a comment why we are adding the rawxml_ query param here.

jpmcmu (changes requested):

wsclient/src/main/java/org/hpccsystems/ws/client/BaseHPCCWsClient.java
rpastrana: I considered URL, but you were in the midst of removing URL from our project. 
At this point, it doesn't seem worthwhile to introduce URL here to add a query param

rpastrana: Looked into the URL/URI classes since I'm adding the comment your suggested, but didn't find any convenient methods allowing us to easily add a query param

rpastrana:
@jpmcmu please review

jpmcmu (approved):
Looks good

PR #716, opened 2024-06-14
HPCC4J-611 Add OpenTelemetry tracing to dfsclient

jpmcmu:
- Added tracing support to read pathways
- Added tracing support to write pathways

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
@@ -27,6 +27,11 @@ import javax.net.ssl.SSLSocket;
 import javax.net.ssl.SSLSocketFactory;
 
 import org.json.JSONObject;
+
+import io.opentelemetry.api.common.Attributes;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.semconv.ServerAttributes;
+
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 import org.hpccsystems.commons.ecl.RecordDefinitionTranslator;
@@ -61,6 +66,9 @@ public class RowServiceOutputStream extends OutputStream
     private long                 bytesWritten                  = 0;
     private long                 handle                        = -1;
     private ByteBuffer           scratchBuffer                 = ByteBuffer.allocate(SCRATCH_BUFFER_LEN);
+
+    private Span                 writeSpan                     = null;
+    private String               traceContextHeader            = null;
 
     private static class RowServiceResponse
     {
@@ -154,7 +162,7 @@ public class RowServiceOutputStream extends OutputStream
     RowServiceOutputStream(String ip, int port, boolean useSSL, String accessToken, FieldDef recordDef, int filePartIndex, String filePartPath,
             CompressionAlgorithm fileCompression, int connectTimeoutMs) throws Exception
     {
-        this(ip,port,useSSL,accessToken,recordDef,filePartIndex,filePartPath,fileCompression, connectTimeoutMs, DEFAULT_SOCKET_OP_TIMEOUT_MS);
+        this(ip,port,useSSL,accessToken,recordDef,filePartIndex,filePartPath,fileCompression, connectTimeoutMs, DEFAULT_SOCKET_OP_TIMEOUT_MS, null);
     }
 
     /**
@@ -186,6 +194,40 @@ public class RowServiceOutputStream extends OutputStream
     RowServiceOutputStream(String ip, int port, boolean useSSL, String accessToken, FieldDef recordDef, int filePartIndex, String filePartPath,
             CompressionAlgorithm fileCompression, int connectTimeoutMs, int sockOpTimeoutMS) throws Exception
     {
+        this(ip,port,useSSL,accessToken,recordDef,filePartIndex,filePartPath,fileCompression, connectTimeoutMs, sockOpTimeoutMS, null);
+    }
+
+    /**
+     * Creates RowServiceOutputStream to be used to stream data to target dafilesrv on HPCC cluster.
+     *
+     * @param ip
+     *            the ip
+     * @param port
+     *            the port
+     * @param useSSL
+     *            the use SSL
+     * @param accessToken
+     *            the access token
+     * @param recordDef
+     *            the record def
+     * @param filePartIndex
+     *            the file part index
+     * @param filePartPath
+     *            the file part path
+     * @param fileCompression
+     *            the file compression
+     * @param connectTimeoutMs
+     *            the socket connect timeout in ms (default is 5000)
+     * @param socketOpTimeoutMS
+     *            the socket operation(read/write) timeout in ms (default is 15000)
+     * @param writeSpan
+     *            the opentelemetry span to use for tracing
+     * @throws Exception
+     *             the exception
+     */
+    RowServiceOutputStream(String ip, int port, boolean useSSL, String accessToken, FieldDef recordDef, int filePartIndex, String filePartPath,
+            CompressionAlgorithm fileCompression, int connectTimeoutMs, int sockOpTimeoutMS, Span writeSpan) throws Exception
+    {
         this.rowServiceIP = ip;
         this.rowServicePort = port;
         this.recordDef = recordDef;
@@ -194,6 +236,19 @@ public class RowServiceOutputStream extends OutputStream
         this.accessToken = accessToken;
         this.compressionAlgo = fileCompression;
         this.sockOpTimeoutMs = sockOpTimeoutMS;
+
+        if (writeSpan != null && writeSpan.getSpanContext().isValid())
+        {
+            this.writeSpan = writeSpan;
+            this.traceContextHeader = org.hpccsystems.ws.client.utils.Utils.getTraceParentHeader(writeSpan);
+        }
+
+        if (this.writeSpan != null)
+        {
+            Attributes attributes = Attributes.of(ServerAttributes.SERVER_ADDRESS, this.rowServiceIP,
+                                                  ServerAttributes.SERVER_PORT, Long.valueOf(this.rowServicePort));
+            writeSpan.addEvent("RowServiceOutputStream.connect", attributes);
+        }
 
         try
         {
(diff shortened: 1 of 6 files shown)


dfsclient/src/main/java/org/hpccsystems/dfs/client/Utils.java
```diff
@@ -0,0 +1,43 @@
...
+import io.opentelemetry.context.Context;
+
+public class Utils
+{
+    public static Tracer getTelemetryTracer()
+    {
+        return GlobalOpenTelemetry.get().getTracer("DFSClient");
+    }
+
+    public static Span createRootSpan(String name)
```
rpastrana, line 28: I'm not sure DFUClient should ever create a root span, since this is a client library we should inherit the trace from the caller

rpastrana (commented):
@jpmcmu provided some high level feedback, let's discuss offline

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -254,7 +320,15 @@ private static class RowServiceResponse
         }
         catch (IOException e)
         {
-            throw new HpccFileException("Failed on initial remote read read trans", e);
+            HpccFileException wrappedException = new HpccFileException("Failed on initial remote read read trans", e);
+            if (writeSpan != null)
+            {
+                Attributes attributes = Attributes.of(ServerAttributes.SERVER_ADDRESS, this.rowServiceIP,
```
rpastrana, line 326: I noticed this seems to be repeated in a few places, why not add a function in utils to handle the attributes annotation in a single place. Also, why aren't we annotating the span as soon as it's created?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -287,16 +369,19 @@ private static class RowServiceResponse
...
+        final String trace = writeSpan != null ? "\"_trace/traceparent\" : \"" + writeSpanTraceID + "\",\n" : "";
+        final String versionMsg = RFCCodes.RFCStreamReadCmd + "{ \"command\" : \"version\", \"handle\": \"-1\", " + trace + " \"format\": \"binary\" }";
         return versionMsg;
     }
 
     private void makeInitialWriteRequest() throws Exception
     {
         String jsonRecordDef = RecordDefinitionTranslator.toJsonRecord(this.recordDef).toString();
 
+        final String trace = writeSpan != null ? "\"_trace/traceparent\" : \"" + writeSpanTraceID + "\",\n" : "";
```
rpastrana, line 381: we might not want to report traceparents unless we know it's a valid traceparent

rpastrana (commented):
@jpmcmu looks fine, I did leave a couple of comments we need to resolve. 

I don't want to delay this pr anymore than needed, but we have to confirm the traceparent is being propagated in the dafilserv requests (perhaps simple junit test cases?)

Also,  since you're creating new trace/spans, we need to exercise that logic as well. 
Can you capture log output showing the toplevel trace/span, and at least one of the child spans and confirm the child inherits the trace id from the parent span?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: I want to cleanup how I am handling adding events / exceptions in the next PR, so I was holding off on cleaning up how attributes get added for those events. However, those attributes should have been added to the span when it was originally created. I will do that as part of this PR.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jpmcmu: Good point, I added a check when during construction, such that writeSpan will only be set if its context is valid.

jpmcmu:
@rpastrana Still looking into unit tests to capture the trace IDs, but here is example log output for a read: 
[otel.javaagent 2024-07-03 15:28:52:230 -0400] [Thread-2] INFO io.opentelemetry.exporter.logging.LoggingSpanExporter - 'HPCCRemoteFileReader.RowService/Read_benchmark::integer::20kb_1' : ac3aaf1874cbf5569e33d679558bc44d e9b52d6c621e6f38 CLIENT [tracer: DFSClient:] AttributesMap{data={server.address=192.168.4.41, server.port=7601, thread.id=18, read.size=-1000, thread.name=Thread-2}, capacity=128, totalAddedValues=5}

rpastrana (approved):
@jpmcmu approved

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -287,16 +374,19 @@ private static class RowServiceResponse
 
     private String makeGetVersionRequest()
     {
-        final String versionMsg = RFCCodes.RFCStreamReadCmd + "{ \"command\" : \"version\", \"handle\": \"-1\", \"format\": \"binary\" }";
+        final String trace = traceContextHeader != null ? "\"_trace/traceparent\" : \"" + traceContextHeader + "\",\n" : "";
```
rpastrana, line 377: @jakesmith FYI all HPCC4J requests into DAFILESrv will contain a new header with OTel trace context. We've confirmed on our end that the header does not seem to break the server side, but we'd like to get your thoughts. Also, we're opening a jira requesting that dafilesrv at minimum consume this info and propagate it, and optimally add helpful instrumentation based on jtrace.

jakesmith: @jpmcmu - is a backslash in an JSON identifier valid? Regardless it looks a little odd, why not just "traceparent" or if want a separate node, generate '"_trace" : { "traceparent": traceContextHeader }' ?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
```diff
@@ -513,8 +662,11 @@ public void write(byte[] b) throws IOException
      */
     public void write(byte[] b, int off, int len) throws IOException
     {
-        String request = "{ \"format\" : \"binary\", \"handle\" : \"" + this.handle + "\""
-                       + (useOldProtocol ? "" : ", \"command\" : \"continue\"") + " }";
+        final String trace = traceContextHeader != null ? "\"_trace/traceparent\" : \"" + traceContextHeader + "\",\n" : "";
```
jakesmith, line 665: is there any value in sending this every request ?
Will it always match the one sent in makeInitialWriteRequest ? If so, dafilesrv can maintain what it saw and reuse, rather than re-read etc.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
rpastrana: @jakesmith it's supposed to be 
```
"_trace": { "traceparent": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00" } ,
```
as you point out. Once fixed, will that layout work for your side?

jakesmith: Personally, I would make it :
"trace": { "parent": "00-4bf92f3577b34da6a3ce929d0e0e4736-00f067aa0ba902b7-00" }

rpastrana: The leading underscore comes from the roxie protocol, I vote we keep it for symetry unless it breaks some sort of convention

jakesmith: right ok, and "traceparent" is also part of that protocol?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
rpastrana: I think the most impactful value in reporting it is validation of expected parent trace. The server side can choose to drop it and ignore, how concerned are we about the potentially superfluous header in that case?

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
rpastrana: yes, it was recently added: https://hpccsystems.atlassian.net/issues/HPCC-31685

jakesmith: ok. So bar the fix to the format, it seems fine.

dfsclient/src/main/java/org/hpccsystems/dfs/client/RowServiceOutputStream.java
jakesmith: so the expectation is that it should match the initial request value? If it does not, should it ignore it and warn, or use the new value. Is it likely to change, or is it impossible?

In general, the dafilesrv protocol does not expect clients to resend info. that has been encompassed in the initial request, e.g. the "node" type and format etc. - the "handle" on all subsequent requests represents the reference to established stream info.

If this is a field value that might change, then I'm not particularly concerned with the extra expense of the comm traffic, assuming the client is dealing with reasonable sized chunks.

rpastrana: it comes down to how much overhead we're incurring vs how much it could help us track a problematic activity...
The overhead seems minimal, but it also doesn't seem particularly helpful.

If the client/server ever get out of synch (I don't know how likely this scenario is), could the traceparent help us track the problem (ie, the out of synch request would be tracked under a known trace)?

jakesmith: I'm not particularly fussed either way, but it is out of character with other info that is sent once.
And, no, I don't think it will be useful if the client gets out of sync. If that happens, the handle (which is the key reference), will be out of sync, the server (dafilesrv) will reject it (challenge response) and the client should reinitiate the connection, along with sending "_trace" info.

PR #756, opened 2024-09-16
HPCC4J-644 FileUtilityTest atomic record count false sharing

jpmcmu:
- Fixed false sharing on atomic counter in FileUtility

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
@@ -843,11 +843,13 @@ public class FileUtility
                         readContext.originalRD = recordDef;
                         HpccRemoteFileReader<HPCCRecord> fileReader = new HpccRemoteFileReader<HPCCRecord>(readContext, filePart, new HPCCRecordBuilder(recordDef));
 
+                        long recCount = 0;
                         while (fileReader.hasNext())
                         {
                             HPCCRecord record = fileReader.next();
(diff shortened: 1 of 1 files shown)


jpmcmu:
@ghalliday @mckellyln  Atomic counter fix. Note: FileUtility is a debugging utility, doesn't affect end user code

rpastrana (approved):
@jpmcmu this looks like a sensible fix based on the offline explanation you provided on the atomic increment issue.

jpmcmu:
Background information on the issue regarding the atomics:

Atomics are fast as long as the cache line that is storing that atomic is in the CPU core's local cache, but when running with a bunch of threads each CPU core first has to request that cache line be moved to their local core's cache before it can be incremented. The end result is each thread is waiting around for that cache line to be moved their core so they can increment it. The result of this false sharing is a problem a lot like lock contention even though atomics are lockless.

rpastrana:
> Background information on the issue regarding the atomics:
> 
> Atomics are fast as long as the cache line that is storing that atomic is in the CPU core's local cache, but when running with a bunch of threads each CPU core first has to request that cache line be moved to their local core's cache before it can be incremented. The end result is each thread is waiting around for that cache line to be moved their core so they can increment it. The result of this false sharing is a problem a lot like lock contention even though atomics are lockless.

Thanks for tracking this down @jpmcmu

jpmcmu:
@rpastrana retargeted to 9.6.x

PR #758, opened 2024-09-17
HPCC4J-647 FileUtility only use otel autosdk without javaagent

jpmcmu:
- Added a check to only initialize the OTel autosdk if an OTel javaagent hasn't been set

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
@@ -75,7 +75,7 @@ public class FileUtility
     private static final int NUM_DEFAULT_THREADS = 4;
     static private final int DEFAULT_ACCESS_EXPIRY_SECONDS = 120;
 
-    private static boolean otelInitialized = false;
+    private static boolean otelNeedsInit = true;
 
     private static class TaskContext
     {
@@ -1973,7 +1973,7 @@ public class FileUtility
      */
     public static JSONArray run(String[] args)
     {
-        if (!otelInitialized)
+        if (otelNeedsInit)
         {
             if (Boolean.getBoolean("otel.java.global-autoconfigure.enabled"))
             {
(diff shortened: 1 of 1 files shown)


dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
```diff
@@ -2140,7 +2140,10 @@ public static JSONArray run(String[] args)
                 System.out.println("    otel.metrics.exporter: "+ System.getProperty("otel.metrics.exporter"));
                 System.out.println("    OTEL_METRICS_EXPORTER Env var: " + System.getenv("OTEL_METRICS_EXPORTER"));
 
-                OpenTelemetry otel = AutoConfiguredOpenTelemetrySdk.initialize().getOpenTelemetrySdk();
+                if (!org.hpccsystems.ws.client.utils.Utils.isOtelJavaagentUsed())
+                {
+                    AutoConfiguredOpenTelemetrySdk.initialize().getOpenTelemetrySdk();
```
rpastrana, line 2145: is "otelInitialized" in L:2149 accurate?
It seems to assume otel has been initialized even if we don't know for sure.

rpastrana (approved):
@jpmcmu approved, but I did post a concern

dfsclient/src/main/java/org/hpccsystems/dfs/client/FileUtility.java
jpmcmu: Good question, I am using that to prevent attempting to setup Otel multiple times. (Run can be called more than once.) So, that should still work as intended.

rpastrana: per offline convo let's refactor the variable name.

jpmcmu:
@rpastrana updated var name and retargeted to 9.6.x

PR #804, opened 2025-04-04
HPCC4J-703 DFSClient: Add tracing on deserialization code paths

jpmcmu:
- Added tracing to deserialization code paths / record construction

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization

### DIFF
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
@@ -17,6 +17,7 @@ import java.math.BigDecimal;
 import java.math.BigInteger;
 import java.math.MathContext;
 import java.nio.charset.Charset;
+import java.security.InvalidParameterException;
 import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
@@ -140,6 +141,8 @@ public class BinaryRecordReader implements IRecordReader
     public static final int      TRIM_STRINGS = 1;
     public static final int      TRIM_FIXED_LEN_STRINGS = 2;
     public static final int      CONVERT_EMPTY_STRINGS_TO_NULL = 4;
+
+    // Defaulting to 100MB to match read stream span size of 100MB
     public static final int      RECORD_BATCH_SIZE_IN_BYTES = 100 * 1000 * 1000;
 
     private boolean              shouldTrimFixedLenStrings = false;
@@ -240,6 +243,11 @@ public class BinaryRecordReader implements IRecordReader
         this.inputStream = new CountingInputStream(is);
         this.inputStream.streamPos = streamPos;
         this.recordBuilderSpan = recBuildSpan;
+        if (this.recordBatchSpan != null && !this.recordBatchSpan.getSpanContext().isValid())
+        {
+            this.recordBatchSpan = null;
+        }
+
         this.defaultLE = true;
 
         if (this.inputStream.markSupported() == false)
@@ -255,11 +263,11 @@ public class BinaryRecordReader implements IRecordReader
      * 
      * @param recordBatchSizeInKB the record batch size in KB
      */
-    public void setRecordBuildingSpanBatchSizeKB(int recordBatchSizeInKB)
+    public void setRecordBuildingSpanBatchSizeKB(int recordBatchSizeInKB) throws InvalidParameterException
     {
         if (recordBatchSizeInKB <= 0)
         {
-            return;
+            throw new InvalidParameterException("BinaryRecordReader.setRecordBuildingSpanBatchSizeKB(): recordBatchSizeInKB must be greater than 0");
         }
 
         this.recordBatchSize = recordBatchSizeInKB * 1024;
--- a/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
+++ b/dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
@@ -22,6 +22,9 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.NoSuchElementException;
 
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.api.trace.StatusCode;
+
 import org.apache.logging.log4j.Logger;
 import org.apache.logging.log4j.LogManager;
 import org.hpccsystems.commons.ecl.FieldDef;
@@ -137,10 +140,23 @@ public class BinaryRecordReader implements IRecordReader
     public static final int      TRIM_STRINGS = 1;
     public static final int      TRIM_FIXED_LEN_STRINGS = 2;
     public static final int      CONVERT_EMPTY_STRINGS_TO_NULL = 4;
+    public static final long     DEFAULT_RECORD_BATCH_SIZE = 10000;
 
     private boolean              shouldTrimFixedLenStrings = false;
     private boolean              shouldTrimStrings = false;
     private boolean              convertEmptyStringsToNull = false;
+
+    private Span                 recordBuilderSpan = null;
+    private Span                 recordBatchSpan = null;
+
+    private long                 recordDeserialTimeNS = 0;
+    private long                 lastRecordTimeNS = 0;
+    private long                 externalProcessingTimeNS = 0;
+    private long                 recordBatchStart = 0;
+    private long                 recordBatchSize = DEFAULT_RECORD_BATCH_SIZE;
+
+    private int                  readStalls = 0;
+    private long                 recordCount = 0;
 
     private byte[]               scratchBuffer = new byte[BUFFER_GROW_SIZE];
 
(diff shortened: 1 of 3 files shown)


dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -216,6 +246,29 @@ public BinaryRecordReader(InputStream is, long streamPos) throws Exception
...
+            if (this.recordBatchSpan != null)
+            {
+                this.recordBatchSpan.setAttribute("RecordCount", (this.recordCount - this.recordBatchStart));
+                this.recordBatchSpan.setAttribute("DeserializationTimeMS", (this.recordDeserialTimeNS / 1000.0f));
+                this.recordBatchSpan.setAttribute("ExternalProcessingTimeMS", (this.externalProcessingTimeNS / 1000.0f));
+                this.recordBatchSpan.updateName( "RecordBatch[" + this.recordBatchStart + "," + this.recordCount + "]");
+                this.recordBatchSpan.end();
+            }
+
+            this.recordBatchSpan = org.hpccsystems.dfs.client.Utils.createChildSpan(recordBuilderSpan, "RecordBatch[]");
```
rpastrana, line 262: does this overwrite the actions in the above block?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -216,6 +246,29 @@ public BinaryRecordReader(InputStream is, long streamPos) throws Exception
...
+    {
+        if (this.recordBuilderSpan != null)
+        {
+            if (this.recordBatchSpan != null)
+            {
+                this.recordBatchSpan.setAttribute("RecordCount", (this.recordCount - this.recordBatchStart));
+                this.recordBatchSpan.setAttribute("DeserializationTimeMS", (this.recordDeserialTimeNS / 1000.0f));
+                this.recordBatchSpan.setAttribute("ExternalProcessingTimeMS", (this.externalProcessingTimeNS / 1000.0f));
+                this.recordBatchSpan.updateName( "RecordBatch[" + this.recordBatchStart + "," + this.recordCount + "]");
+                this.recordBatchSpan.end();
```
rpastrana, line 259: it's odd to me to end a member span in a method named startNewRecordBatchSpan. Are we closing out a previous span and starting a brand new one?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -216,6 +246,29 @@ public BinaryRecordReader(InputStream is, long streamPos) throws Exception
...
+    private void startNewRecordBatchSpan()
+    {
+        if (this.recordBuilderSpan != null)
+        {
+            if (this.recordBatchSpan != null)
+            {
+                this.recordBatchSpan.setAttribute("RecordCount", (this.recordCount - this.recordBatchStart));
+                this.recordBatchSpan.setAttribute("DeserializationTimeMS", (this.recordDeserialTimeNS / 1000.0f));
+                this.recordBatchSpan.setAttribute("ExternalProcessingTimeMS", (this.externalProcessingTimeNS / 1000.0f));
+                this.recordBatchSpan.updateName( "RecordBatch[" + this.recordBatchStart + "," + this.recordCount + "]");
```
rpastrana, line 258: are there other paths (exceptions) that would end the span under the original name?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -318,6 +371,9 @@ public boolean hasNext() throws HpccFileException
      */
     public Object getNext() throws HpccFileException
     {
+        long start = System.nanoTime();
```
rpastrana, line 374: can we avoid calculating span attributes if not tracing in an efficient manner?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -773,6 +843,7 @@ private void readIntoScratchBuffer(int offset, int dataLen) throws IOException
 
         if (totalSleepTimeMS > SLEEP_TIME_WARN_MS)
         {
+            this.readStalls++;
```
rpastrana, line 846: is this to be tracked in one of the spans?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -208,6 +237,7 @@ public BinaryRecordReader(InputStream is, long streamPos) throws Exception
 
         this.inputStream = new CountingInputStream(is);
         this.inputStream.streamPos = streamPos;
+        this.recordBuilderSpan = recBuildSpan;
```
rpastrana, line 240: I think recordBuilderSpan is acting as a parent span for all potential recordBatchSpans, is it tracking anything else? and why is it passed in, what is the caller tracking in this span

rpastrana (changes requested):

jpmcmu:
@rpastrana Pushed up code review changes, and made changes to sync the read request span size and record building span size by default to make it easier to see issues.
Screenshot 2025-04-17 at 9 26 31 AM

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
jpmcmu: Good question, I added a check to close out the span when the input stream is closed and or when we are out of records.

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -244,26 +246,52 @@ public BinaryRecordReader(InputStream is, long streamPos, Span recBuildSpan) thr
...
+        {
+            return;
+        }
+
+        this.recordBatchSize = recordBatchSizeInKB * 1024;
+    }
+
+    private void endRecordBatchSpan()
+    {
+        if (this.recordBatchSpan == null)
```
rpastrana, line 270: is it possible for recordBatchSpan != null, but be invalid/not recording?  if so, could we skip the stop process?

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
```diff
@@ -371,8 +406,12 @@ public boolean hasNext() throws HpccFileException
      */
     public Object getNext() throws HpccFileException
     {
-        long start = System.nanoTime();
-        this.externalProcessingTimeNS += (start - lastRecordTimeNS);
+        long start = 0;
+        if (this.recordBuilderSpan != null)
```
rpastrana, line 410: Have we established a simple way determine if tracing is currently not enabled? I'm concerned that this.recordBuilderSpan could be instantiated, but not valid/recording. In those situations, we should avoid gathering timing information for performance reasons.

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
```diff
@@ -75,6 +75,24 @@ public static class FileReadContext
         public int readBufferSizeKB = -1;
         public int readRequestSpanBatchSize = -1; // The number of read requests before creating a new span
         public Span parentSpan = null;
+
+        private long getReadSizeKB()
+        {
+            if (readSizeKB <= 0)
```
rpastrana, line 81: should we ever allow readSizeKB to be <=0 ?

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
```diff
@@ -416,6 +440,7 @@ private boolean retryRead()
                 this.inputStream.skip(bytesToSkip);
 
                 this.binaryRecordReader = new BinaryRecordReader(this.inputStream, resumeInfo.recordReaderStreamPos, this.readSpan);
+                this.binaryRecordReader.setRecordBuildingSpanBatchSizeKB(context.getReadRequestSpanBatchSize() * (int) context.getReadSizeKB());
```
rpastrana, line 443: should we be concerned about the potential data loss in the long to int cast? if not, does getReadSizeKB need to return a long?

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
jpmcmu: In this case we are treating -1 as a "use the default value" indicator so we need to check for < 0 and default

dfsclient/src/main/java/org/hpccsystems/dfs/client/HpccRemoteFileReader.java
jpmcmu: I think the 32bit range should be good here, it would allow for 2TB span sizes due to counting KB not bytes. I don't see any use case for having spans representing that much data.

dfsclient/src/main/java/org/hpccsystems/dfs/client/BinaryRecordReader.java
jpmcmu: Good question, I added a check when setting the initial span in the BinaryRecordReader to null out the span if it isn't valid.

PR #836, opened 2025-05-15
HPPCC4J-708 Synchronize with HPCC 9.12.x ESP interfaces

rpastrana:
- Updates all services to match HPCC 9.12.x
- Updates all services to match HPCC 9.12.x
- Updates pom to match filespray wsdl naming
- Updates vscode launch to include several individual service updaters

## Type of change:

## Checklist:
  - [ ] Scalability
  - [ ] Performance
  - [ ] Security
  - [ ] Thread-safety
  - [ ] Premature optimization


.vscode/launch.json
```diff
@@ -4,6 +4,16 @@
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
```
jpmcmu, line 4: Does the comment type work in JSON?

jpmcmu (approved):
@rpastrana Minor question, but otherwise looks good to me. As an aside would it make sense to add some documentation on how to use the launch configurations to update the interfaces to our wiki?

.vscode/launch.json
rpastrana: vscode seems to handle the comment without issue, do we think it should be removed?

jpmcmu: Ah ok, in that case I think it is fine

rpastrana:
@jpmcmu please review latest commit

wsclient/src/main/resources/WSDLs/WsWorkunits-200.wsdl.notes
```diff
@@ -0,0 +1,6 @@
+WsWorkunits 2.00 required manual changes to WSDL to force successful compilation.
```
jpmcmu, line 1: WsWorkunits 2.00 doesn't exist on the ESP side right? Will there be a conflict if / when ESP moves to 2.00?

jpmcmu (commented):
@rpastrana Looks like javadocs are still failing and a minor question on the new 2.0 version, otherwise looks good.

rpastrana:
@jpmcmu please take a look at the 2 most recent commits.

jpmcmu (approved):
Minor question about commented out code, otherwise looks good

